PR #52, opened 2025-03-27
feat: add reporter `onTestRunEnd` hook

9aoy:
## Summary

- add reporter `onTestRunEnd` hook.It will be called after all tests have finished running.
   - printSummaryLog when `onTestRunEnd` calls.
- rename `onTestEnd` to `onTestCaseResult`.  Align with `vitest` and `jest`.
- add prepare time into build.

image

## Checklist

### DIFF
--- a/packages/core/src/reporter/summary.ts
+++ b/packages/core/src/reporter/summary.ts
@@ -42,7 +42,7 @@ export const printSummaryLog = (
   );
 
   logger.log(
-    `${color.gray('Duration'.padStart(12))} ${prettyTime(duration.totalTime)} ${color.gray(`(build ${prettyTime(duration.buildTime)}, tests ${prettyTime(duration.testTime)}, prepare ${prettyTime(duration.prepareTime)}`)})`,
+    `${color.gray('Duration'.padStart(12))} ${prettyTime(duration.totalTime)} ${color.gray(`(build ${prettyTime(duration.buildTime)}, tests ${prettyTime(duration.testTime)}`)})`,
   );
   logger.log('');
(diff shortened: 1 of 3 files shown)


packages/core/src/reporter/summary.ts
```diff
@@ -31,22 +31,18 @@ export const getStatusString = (
...
+    `${color.gray('Test Files'.padStart(12))} ${getSummaryStatusString(results)}`,
   );
   logger.log(
-    `${color.gray('Tests'.padStart(12))} ${getStatusString(testResults)}`,
+    `${color.gray('Tests'.padStart(12))} ${getSummaryStatusString(testResults)}`,
   );
 
   logger.log(
-    `${color.gray('Duration'.padStart(12))} ${prettyTime(duration.totalTime)} ${color.gray(`(build ${prettyTime(duration.buildTime)}, tests ${prettyTime(duration.testTime)}`)})`,
+    `${color.gray('Duration'.padStart(12))} ${prettyTime(duration.totalTime)} ${color.gray(`(build ${prettyTime(duration.buildTime)}, tests ${prettyTime(duration.testTime)}, prepare ${prettyTime(duration.prepareTime)}`)})`,
```
fi3ework: I think it's better to put `prepare` at first as it happens ahead.

9aoy: But I think the time cost on `prepare` is not particularly necessary for users.

fi3ework: Remove or merge it into build is also fine.

PR #62, opened 2025-03-31
fix(reporter): print posix file path

9aoy:
## Summary

should print posix file path in windows.
image

## Checklist

### DIFF
--- a/packages/core/src/reporter/index.ts
+++ b/packages/core/src/reporter/index.ts
@@ -1,4 +1,4 @@
-import { parse, relative } from 'node:path';
+import { posix } from 'node:path';
 import type {
   Duration,
   GetSourcemap,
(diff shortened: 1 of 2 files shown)


packages/core/src/reporter/index.ts
```diff
@@ -23,7 +23,9 @@ export class DefaultReporter implements Reporter {
     const { dir, base } = parse(relativePath);
 
     console.log('');
-    console.log(`${color.gray(`> ${dir ? `${dir}/` : ''}`)}${base}`);
+    console.log(
+      `${color.gray(`> ${dir ? `${dir.split(sep).join('/')}/` : ''}`)}${base}`,
```
fi3ework: I'm afraid it will be used quite often, what if extract it into a util function?

fi3ework (approved):

PR #63, opened 2025-03-31
feat: support basic watch mode

fi3ework:
## Summary

Support basic watch mode, implement with context dependency.

## Checklist

### DIFF
--- a/examples/node/package.json
+++ b/examples/node/package.json
@@ -4,7 +4,8 @@
   "type": "commonjs",
   "scripts": {
     "build": "rsbuild build",
-    "test": "rstest run"
+    "test": "rstest run",
+    "test:watch": "rstest"
   },
   "devDependencies": {
     "@rstest/core": "workspace:*",
--- a/packages/core/src/cli/commands.ts
+++ b/packages/core/src/cli/commands.ts
@@ -82,16 +82,18 @@ export function setupCommands(): void {
   applyCommonOptions(cli);
 
   cli
-    .command('[...filters]', 'run Rstest')
+    .command('[...filters]', 'run tests in watch mode')
     .action(async (filters: string[], options: CommonOptions) => {
       try {
         const { config } = await initCli(options);
         const { createRstest } = await import('../core');
-        if (isCI) {
+        // TODO: remove RSTEST_SELF_CI when we can run test without CLI.
+        if (isCI && !RSTEST_SELF_CI) {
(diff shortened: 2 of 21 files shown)


packages/core/src/cli/commands.ts
```diff
@@ -82,17 +82,19 @@ export function setupCommands(): void {
   applyCommonOptions(cli);
 
   cli
-    .command('[...filters]', 'run Rstest')
+    .command('[...filters]', 'run tests in watch mode')
     .action(async (filters: string[], options: CommonOptions) => {
       try {
         const { config } = await initCli(options);
         const { createRstest } = await import('../core');
-        if (isCI) {
```
fi3ework: Do we need to auto turn off `watch` mode in CI?

packages/core/src/core/runTests.ts
```diff
@@ -14,70 +14,98 @@ export async function runTests(
...
+      sourceMaps,
+      setupEntries,
+      assetFiles,
+      context,
     });
+    const testEnd = Date.now();
 
-  const buildEnd = Date.now();
+    const duration = {
+      totalTime: testEnd - start,
```
9aoy, line 80: shell we need reset start time when watch rerun?

packages/core/src/cli/commands.ts
fi3ework: Or we need an other flag as an escape to allow run watch in Rstest CI test.

9aoy: `rstest watch` will always run watch. `rstest` will enter the watch mode in development environment and run mode in CI automatically.

packages/core/src/core/runTests.ts
fi3ework: Nice catch.

packages/core/src/cli/commands.ts
fi3ework: Use a build phase define to escape as a workaround.

examples/node/package.json
```diff
@@ -4,7 +4,8 @@
   "type": "commonjs",
   "scripts": {
     "build": "rsbuild build",
-    "test": "rstest run"
+    "test": "rstest run",
+    "test:watch": "rstest"
```
9aoy, line 8: why not use `rstest watch` command?

image

fi3ework: yeah, fixed!

9aoy (approved):

PR #76, opened 2025-04-08
feat: support `toMatchInlineSnapshot` API

9aoy:
Support `toMatchInlineSnapshot` API. The inline snapshot should run code in `dist` and save snapshot in `source`.

* Add `RstestSnapshotEnvironment` to handle source maps for inline snapshots, ensuring accurate stack traces (StackTrace is used to parse inline snapshot position)
* Added `keep_fnames` minification options to the `rslib.config.ts` file to preserve function names for stack traces

image

## Checklist

### DIFF
--- a/packages/core/src/worker/snapshot.ts
+++ b/packages/core/src/worker/snapshot.ts
@@ -1,0 +1,46 @@
+import { TraceMap, originalPositionFor } from '@jridgewell/trace-mapping';
+import { NodeSnapshotEnvironment } from '@vitest/snapshot/environment';
+import type { SourceMapInput } from '../types';
+
+interface ParsedStack {
+  method: string;
+  file: string;
+  line: number;
+  column: number;
+}
+
+export class RstestSnapshotEnvironment extends NodeSnapshotEnvironment {
+  private sourceMaps: Record<string, SourceMapInput> = {};
+
+  constructor({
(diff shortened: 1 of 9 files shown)


packages/core/src/worker/snapshot.ts
```diff
@@ -0,0 +1,46 @@
...
+    sourceMaps,
+  }: {
+    sourceMaps: Record<string, SourceMapInput>;
+  }) {
+    super();
+    this.sourceMaps = sourceMaps;
+  }
+
+  // StackTrace is used to parse inline snapshot position
+  // The inline snapshot should run code in dist and save snapshot in source
```
fi3ework, line 25: I guess `'-u'` mode is not finished so original source position is not used yet?

9aoy: `-u` feature has been supported, but the test has not been added yet. 😢  https://github.com/web-infra-dev/rstest/pull/69

And source position is used in two scenarios in snapshot:
- snapshot comparison failed, prompting error source code information
- inline snapshot first generated / updated.

fi3ework: Got it, I'm working on watch mode test and will add more test util functions which could support update source file in a temp dir. We could add the test for `-u` after that.

PR #82, opened 2025-04-10
feat: support register async suite

9aoy:
## Summary

Support register **async** suite. BTW, If you have some preparations that need to be executed before testing, it is recommended to use `beforeEach` / `beforeAll`.

```ts
  describe('Test Async Suite', async () => {
    // do something async...
    await new Promise<void>((resolve) => {
      setTimeout(() => {
        resolve();
      }, 100);
    });

   it('should add two numbers correctly', () => {
      expect(1 + 1).toBe(2);
    });
  });
```
## Checklist

### DIFF
--- a/tests/runner/test/fixtures/async.test.ts
+++ b/tests/runner/test/fixtures/async.test.ts
@@ -1,0 +1,66 @@
+import { describe, expect, it } from '@rstest/core';
+
+describe('Test Async Suite', () => {
+  describe('0', async () => {
+    await new Promise<void>((resolve) => {
+      setTimeout(() => {
+        resolve();
+      }, 100);
+    });
+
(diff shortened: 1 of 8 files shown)


tests/runner/test/fixtures/async.test.ts
```diff
@@ -0,0 +1,66 @@
+import { describe, expect, it } from '@rstest/core';
+
+describe('Test Async Suite', () => {
+  describe('0', async () => {
+    await new Promise<void>((resolve) => {
```
fi3ework: Could move the sleep function to utils, we already have 4 similar functions (search `await new Promise`).

fi3ework (approved):

PR #88, opened 2025-04-14
feat: support print diff

9aoy:
## Summary

support print error diff, powered by [jest-diff](https://www.npmjs.com/package/jest-diff) 🌹 .

diff string:
image

diff object:

image

## Checklist

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -46,7 +46,6 @@
     "@rsbuild/core": "^1.3.4",
     "birpc": "2.3.0",
     "chai": "^5.2.0",
-    "jest-diff": "^29.7.0",
     "pathe": "^2.0.3",
     "std-env": "^3.9.0",
     "tinypool": "^1.0.2"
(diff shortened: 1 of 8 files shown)


packages/core/package.json
```diff
@@ -46,6 +46,7 @@
     "@rsbuild/core": "^1.3.4",
     "birpc": "2.3.0",
     "chai": "^5.2.0",
+    "jest-diff": "^29.7.0",
```
fi3ework: `jest-diff`'s dependency is a bit messy https://npmgraph.js.org/?q=jest-diff and it hasn't been maintained by two years (so does Jest). Do we have any alternatives? 👀
image

9aoy: image

29.7.0 is the latest version of `jest`. `jest-diff` is used by jest officially and is relatively reliable and stable. And it seems that jest has been preparing for 30.0.0.

fi3ework: Not so much maintenance but I think the package is relativly stable https://github.com/jestjs/jest/commits/main/packages/jest-diff. The concern is the bloat of indirect dependence. It's not the most important issue for us now, we might change it in the future.

fi3ework (approved):

PR #93, opened 2025-04-14
feat: external `node_modules` packages  by default

9aoy:
## Summary

Externalize `node_modules` packages by default. These dependencies will not be transform & bundled.

## Checklist


packages/core/src/core/rsbuild.ts
```diff
@@ -19,6 +19,47 @@ const isMultiCompiler = <
...
+    return callback();
+  }
+
+  resolver(context!, request!, (err, resolvePath) => {
+    if (err) {
+      // ignore resolve error
+      return callback();
+    }
+
+    if (resolvePath && /node_modules/.test(resolvePath!)) {
```
fi3ework: We could put `/node_modules/.test` logic into `doExternal` to convergence the logic.

9aoy: I think `doExternal` is more general now, it can be irrelevant to `node_modules`.

fi3ework: Never mind, I just saw two same logic to test `node_modules`.

PR #143, opened 2025-04-27
fix: update node engine range to >=18.0

9aoy:
## Summary

reference: https://github.com/web-infra-dev/rsbuild/pull/5032

## Checklist

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -67,7 +67,7 @@
     "tinyspy": "^4.0.3"
   },
   "engines": {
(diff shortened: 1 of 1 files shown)


packages/core/package.json
```diff
@@ -67,7 +67,7 @@
     "tinyspy": "^4.0.3"
   },
   "engines": {
-    "node": ">=16.7.0"
+    "node": ">=16.10.0"
```
chenjiahan, line 70: How about >= 18.0.0?

PR #196, opened 2025-05-16
feat(reporter): support render status

9aoy:
## Summary

image

## Checklist

### DIFF
--- a/packages/core/src/reporter/windowedRenderer.ts
+++ b/packages/core/src/reporter/windowedRenderer.ts
@@ -1,0 +1,214 @@
+import type { Writable } from 'node:stream';
+import { stripVTControlCharacters } from 'node:util';
+
+const DEFAULT_RENDER_INTERVAL_MS = 1_000;
+
+const ESC = '\x1B[';
+const CLEAR_LINE = `${ESC}K`;
+const MOVE_CURSOR_ONE_ROW_UP = `${ESC}1A`;
+const SYNC_START = `${ESC}?2026h`;
+const SYNC_END = `${ESC}?2026l`;
+
+interface Options {
+  logger: {
+    outputStream: Writable;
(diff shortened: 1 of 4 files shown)


packages/core/src/reporter/windowedRenderer.ts
```diff
@@ -0,0 +1,214 @@
...
+type StreamType = 'output' | 'error';
+
+/**
+ * This method is modified based on source found in
+ * https://github.com/vitest-dev/vitest/blob/accd2edaeff7aca01392452ff9d2753d4e448ac5/packages/vitest/src/node/reporters/renderers/windowedRenderer.ts
+ *
+ * Renders content of `getWindow` at the bottom of the terminal and
+ * forwards all other intercepted `stdout` and `stderr` logs above it.
+ */
+export class WindowRenderer {
```
AriPerkkio, line 31: Hey @9aoy - excited to see progress of `rstest` and how Vitest's sources are utilized in other packages.

When copying the sources as direct as here, shouldn't the file have original license and copyrights too instead? Similar to https://github.com/vitest-dev/vitest/blob/9e412de354e48edcd96fde9a86557f7e3275592e/packages/browser/src/client/tester/expect/toBeChecked.ts#L1-L16 ?

9aoy: Thank you for pointing that out! We will correct this oversight. Once again, we really appreciate Vitest's contributions to the testing field and the inspiration it has provided for rstest.

fi3ework: Thanks @AriPerkkio for the hinting, we are very grateful for Vitest’s outstanding work. we'll also add a more general declaration to thank Vitest's work in our license file today(which is not as of now).

PR #284, opened 2025-06-16
feat: support `jsdom` test

9aoy:
## Summary

support `jsdom` test.

1. Install dependencies: 

```bash
pnpm i jsdom -D
```

2. Specify `jsdom` as testEnvironment: 

```ts

export default defineConfig({
  testEnvironment: 'jsdom',
});
```

3. Writing tests

```ts
import { render, screen } from '@testing-library/react';
import App from '../src/App';

test('should render App correctly', async () => {
  render(<App />);

  const element = screen.getByText('Rsbuild with React');

  expect(element.tagName).toBe('H1');
});
```

## Checklist


packages/core/src/runtime/worker/env/jsdom.ts
```diff
@@ -0,0 +1,163 @@
...
+    return removeEventListener.apply(this, args);
+  };
+  return () => {
+    window.removeEventListener('error', throwUnhandledError);
+  };
+}
+
+export default (<TestEnvironment>{
+  name: 'jsdom',
+  async setup(global: any, { jsdom = {} }) {
```
fi3ework: What if `TestEnvironment` accept a general type as the setup `args[0]`, and use `any` as default, we could add some types here at least.

fi3ework (approved):

PR #286, opened 2025-06-17
docs: add `include` and `exclude` configs

9aoy:
## Summary

docs: add `include` and `exclude` configs

## Checklist


website/docs/zh/config/test/exclude.mdx
```diff
@@ -0,0 +1,14 @@
+# exclude
```
fi3ework, line 5: When a file both matched by `include` and `exclude`, will it be included finally? What about document it out by using such `:::info` call out.

PR #290, opened 2025-06-18
feat: more mock api

fi3ework:
## Summary

Support most mock APIs.

## Related Links

- [x] `rs.mock`
- [x] `rs.importActual`
- [x] `rs.doMock`
- [x] `rs.importMock`
- [x] `rs.mockRequire`
- [x] `rs.requireActual`
- [x] `rs.doMockRequire`
- [x] `rs.requireMock`
- [x] `rs.resetModules`
- [x] `rs.unmock`
- [x] `rs.doUnmock`

## Checklist


fi3ework:
It's good to be reviewed now. Although there is still more work to do, it can handle most common mock functionalities. This PR will be able to migrate Rslib mock test cases tested by linking them locally on my machine.

packages/core/src/core/plugins/mockRuntime.ts
```diff
@@ -0,0 +1,73 @@
+import type { RsbuildPlugin, Rspack } from '@rsbuild/core';
+
+class MockRuntimeRspackPlugin {
+  apply(compiler: Rspack.Compiler) {
+    const { RuntimeModule } = compiler.webpack;
+    class RetestImportRuntimeModule extends RuntimeModule {
+      constructor() {
+        super('rstest runtime');
```
9aoy, line 8: Finally know where this `rstest runtime` came from 😆 
image

packages/core/src/runtime/api/utilities.ts
```diff
@@ -43,30 +43,41 @@ export const createRstestUtilities: () => RstestUtilities = () => {
       return rstest;
     },
     mock: () => {
-      // TODO
+      // The actual implementation is managed by the built-in Rstest plugin.
```
9aoy, line 46: Maybe we can throw error, if some dynamic case is not processed by Rstest plugin.

fi3ework: That will be the very ideal target, however, the hard part is: you can mock a module in a test entry file while import it in a nested imported source file, and they're isolated in compile time. So we had to throw an error in runtime.

I think this is still feasible with the following setups in the runtime solution:

1. store all mocked module ids in runtime.
2. monkey patch `__webpack_require__` (it can't be customized render by plugin IMO) to diversion mocked module and non found module, throw a user-friendly error for the later case.

fi3ework: Updated, since in runtime, mocked is supposed to register before exec the module, so we don't need to store all mocked IDs, but only need to throw and readable error message.

9aoy (approved):

PR #292, opened 2025-06-19
fix: `Unexpected token 'export'` error in Node.js 18.

9aoy:
## Summary

fix: `Unexpected token 'export'` error in Node.js 18. 

When `import foo from 'foo'`, Rspack resolves to the module in the `module` field, because the `module` field has the highest priority in [`resolve.mainFields`](https://rspack.rs/config/resolve#resolvemainfiles) array. 

However, when Rstest tries to load this modulePath in Node.js v18, we get the following warning and error:

```bash
(node:34734) Warning: To load an ES module, set "type": "module" in the package.json or use the .mjs extension.

SyntaxError: Unexpected token 'export'
```

## Checklist


9aoy:
Is it necessary to support node v18? since it is EOL now.

packages/core/src/core/rsbuild.ts
```diff
@@ -145,6 +145,12 @@ export const prepareRsbuild = async (
                 ...(config.module.parser.javascript || {}),
               };
 
+              config.resolve ??= {};
+              // skip `module` field. ESM module resolved by module field is not always a native ESM module
+              config.resolve.mainFields = config.resolve.mainFields?.filter(
+                (filed) => filed !== 'module',
```
fi3ework: What if there's a legacy package with `module` and `main`, without `exports` field? like https://github.com/arco-design/arco-design/blob/main/package.json.

njzydark: What if the package is pure ES module? And also, if browser mode is supported later, would the logic here need to be changed?

9aoy: Updated to: Skip the `module` field only in the `node` test environment and when Node.js version is 18 or below.

PR #331, opened 2025-06-27
fix: interop lodash correctly

9aoy:
## Summary

Some packages with invalid named exports like `lodash` should be handled correctly.

```js
// lodash/index.js
module.exports = {
  VERSION: '4.17.21';
}
```
enable interopDefault:
```ts
import lodash, { VERSION } from 'lodash';

expect(lodash.VERSION).toBe(VERSION);
```

disable interopDefault:
```ts
import lodash, { VERSION } from 'lodash';
expect(VERSION).toBe(undefined)
expect(lodash.VERSION).toBe('4.7.12')
```

## Checklist


fi3ework:
IMO, is this a deliberate fix for the interop default to prevent user-side errors, which differs from Node.js's default behavior?

9aoy:
> IMO, is this a deliberate fix for the interop default to prevent user-side errors, which differs from Node.js's default behavior?

Node.js will not resolve invalid named exports. This behavior is consistent with bundler when `interopDefault` is enabled.

fi3ework (approved):

PR #341, opened 2025-06-30
chore: auto exit nx console

fi3ework:
## Summary

It's very hard for me to understand why nx doesn't exit automatically by default.

## Checklist


nx.json
```diff
@@ -1,5 +1,8 @@
 {
   "$schema": "./node_modules/nx/schemas/nx-schema.json",
+  "tui": {
+    "autoExit": true
```
fi3ework:
```suggestion
    "enabled": false
```

PR #343, opened 2025-07-01
fix: should run tests correctly when isolate false

9aoy:
## Summary

 By default, modules are isolated between different tests (powered by tinypool).
When `isolate: false`, we should clean setup and rstest module cache manually, this is used to ensure that the setup files and rstest core are re-executed in each test run.

## Checklist


packages/core/src/core/plugins/moduleCacheControl.ts
```diff
@@ -0,0 +1,69 @@
...
+    return;
+  }
+  delete __webpack_module_cache__['@rstest/core'];
+
+  global.setupIds.forEach((id) => {
+    delete __webpack_module_cache__[id];
+  });
+}
+
+global.__clean_rstest_core_cache__ = __clean_rstest_core_cache__;
```
fi3ework: Can we access `__webpack_require__` instead of using `global` to call `__clean_rstest_core_cache__`, which means we could call `__webpack_require__.__clean_rstest_core_cache__` directly. And also, we'd better prefix all Rstset runtime method with `rstest_` (`rstest_clean_core_cache`), just the same as other runtime methods.

9aoy: If we use `__webpack_require__`, how do we call it at runtime like below?
image

fi3ework: Got it.

packages/core/src/core/plugins/moduleCacheControl.ts
```diff
@@ -0,0 +1,69 @@
...
+ * This is used to ensure that the setup files and rstest core are re-executed in each test run
+ *
+ * By default, modules are isolated between different tests (powered by tinypool).
+ */
+export const pluginCacheControl: (setupFiles: string[]) => RsbuildPlugin = (
+  setupFiles: string[],
+) => ({
+  name: 'rstest:cache-control',
+  setup: (api) => {
+    api.transform({ test: setupFiles }, ({ code }) => {
```
9aoy, line 54: Maybe it's better to inject via initFragment?

PR #349, opened 2025-07-02
fix(mock): resetModules should not clear mock

fi3ework:
## Summary

`resetModules` should not clear mocked modules.

## Checklist

### DIFF
--- a/packages/core/src/core/plugins/mockRuntime.ts
+++ b/packages/core/src/core/plugins/mockRuntime.ts
@@ -36,9 +36,10 @@ __webpack_require__.rstest_original_modules = {};
 
 // TODO: Remove "reset_modules" in next Rspack version.
 __webpack_require__.rstest_reset_modules = __webpack_require__.reset_modules = () => {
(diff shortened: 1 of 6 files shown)


packages/core/src/core/plugins/mockRuntime.ts
```diff
@@ -36,7 +36,12 @@ __webpack_require__.rstest_original_modules = {};
 
 // TODO: Remove "reset_modules" in next Rspack version.
 __webpack_require__.rstest_reset_modules = __webpack_require__.reset_modules = () => {
-  __webpack_module_cache__ = {};
+  Object.keys(__webpack_module_cache__).forEach(id => {
+    // Do not reset mocks registry.
+    if (!Object.keys(__webpack_require__.rstest_original_modules).includes(id)) {
```
9aoy: Moving `Object.keys(__webpack_require__.rstest_original_modules)` up may reduce some duplicate calculations.

9aoy: `@rstest/core` should also not be reset.

PR #360, opened 2025-07-04
chore: optimize the display of time duration when it has minutes

9aoy:
## Summary

optimize the display of time duration when it has minutes.

When the time duration is `110000ms`
- before: `1.83m`
- after: `1m 50s`

## Checklist


fi3ework (commented):
Do we need to add a space between larger time frame and lower larger frame (e.g. 1m 10s), which is also comply with the https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DurationFormat/DurationFormat#narrow API.

PR #364, opened 2025-07-04
perf: disable some of rsbuild's default devMiddlewares

9aoy:
## Summary

Disable some of rsbuild's default devMiddlewares that are not used in rstest.

https://rsbuild.rs/config/

## Checklist

### DIFF
--- a/packages/core/src/core/rsbuild.ts
+++ b/packages/core/src/core/rsbuild.ts
@@ -143,6 +143,12 @@ export const prepareRsbuild = async (
         printUrls: false,
         strictPort: false,
         middlewareMode: true,
+        compress: false,
+        cors: false,
+        publicDir: false,
+      },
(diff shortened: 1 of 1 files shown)


packages/core/src/core/rsbuild.ts
```diff
@@ -143,6 +143,12 @@ export const prepareRsbuild = async (
         printUrls: false,
         strictPort: false,
         middlewareMode: true,
+        compress: false,
+        cors: false,
+        publicDir: false,
+      },
+      dev: {
+        hmr: false,
```
fi3ework, line 151: We will need this after supporting `"browser"` mode in the future, we can bring it back then.

9aoy: yep~

PR #374, opened 2025-07-09
fix: unexpected resolution of vue to `vue.cjs.prod.js` when `NODE_ENV=test`

9aoy:
## Summary

Fix unexpected resolution of vue to `vue.cjs.prod.js` when `NODE_ENV=test`.

By default, Rspack treats `NODE_ENV=test` as `production` in some cases (e.g., `conditionNames`). However, frameworks like Vue and React (and tools like Vitest) expect test to follow development-mode behavior.

image

image

image

https://github.com/web-infra-dev/rspack/blob/8671bc16eb74961dcd98a5bb03eae4efce936914/packages/rspack/src/config/defaults.ts#L1069

## Checklist

### DIFF
--- a/tests/vue/fixtures/test/index.test.ts
+++ b/tests/vue/fixtures/test/index.test.ts
@@ -1,6 +1,5 @@
 import { expect, test } from '@rstest/core';
 import { mount } from '@vue/test-utils';
-// @ts-expect-error
 import App from '../src/App.vue';
 
 test('should emit clickApp event when button is clicked', async () => {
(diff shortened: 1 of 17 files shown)


tests/vue/fixtures/test/index.test.ts
```diff
@@ -0,0 +1,10 @@
+import { expect, test } from '@rstest/core';
+import { mount } from '@vue/test-utils';
+// @ts-expect-error
```
fi3ework: We can remove this, for IDE, we could use Vue official extension. For type check, which we're not using now, we could use `vue-tsc`.

fi3ework (approved):

PR #377, opened 2025-07-10
feat: support using package name as `setupFiles` value

9aoy:
## Summary

support using package name as `setupFiles` value.

```ts

export default defineConfig({
  setupFiles: ['jsdom-global/register', 'rstest-canvas-mock'],
});
```

## Checklist

### DIFF
--- a/packages/core/src/utils/testFiles.ts
+++ b/packages/core/src/utils/testFiles.ts
@@ -1,5 +1,6 @@
 import { existsSync } from 'node:fs';
 import fs from 'node:fs/promises';
+import { createRequire } from 'node:module';
 import pathe from 'pathe';
 import { glob } from 'tinyglobby';
 import { castArray, color, getAbsolutePath, parsePosix } from './helper';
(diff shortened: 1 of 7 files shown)


packages/core/src/utils/testFiles.ts
```diff
@@ -98,23 +99,39 @@ export const getTestEntries = async ({
   );
 };
 
+const tryResolve = (request: string, rootPath: string) => {
+  try {
+    const require = createRequire(rootPath);
+    return require.resolve(request, { paths: [rootPath] });
```
fi3ework, line 105: `require.resolve` will try to resolve to a CJS export when resolving a dual package. I think we could use `import.meta.resolve` since we're ESM only and prefer resolving an ESM entry point first.

9aoy: The second `parent` parameter of `import.meta.resolve` is only used when the `--experimental-import-meta-resolve` command flag enabled, we cannot assume that all users will enable it when using rstest.

Here we just use `require.resolve` to determine that it is a valid path, the path to the real resolve is determined by rspack.

fi3ework: Oh, got it.

Issue #398, opened 2025-07-20
[Feature]: coverage provider API for rstest


travzhang:
## Background & Motivation

Test coverage is a critical metric in JavaScript/TypeScript projects. Currently, `rstest` lacks built-in coverage collection and reporting. This RFC proposes adding coverage support via a pluggable provider system, initially supporting Istanbul instrumentation with future plans for V8 and custom providers.

---

## Goals

- Add `coverage` config in `rstest.config.ts`
- Support code instrumentation and runtime coverage collection
- Merge coverage data from multiple workers
- Generate coverage reports (HTML and JSON)
- Provide a CoverageProvider interface for custom implementations

## Configuration Example

```ts
export default defineConfig({
  coverage: {
    enabled: true,
    provider: 'istanbul', // future: 'v8'
  },
});
````

---

## CoverageProvider Interface

```ts
export interface CoverageProvider {
  init(): void;
  collect(): CoverageMap | null;
  generateReports(coverageMap: CoverageMap, options: CoverageOptions): Promise<void>;
  cleanup(): void;
}
```

---

## Instrumentation Implementation Notes

We explored three approaches for Istanbul instrumentation:

| Strategy                                      | Description                              | Status   | Notes                                                   |
| --------------------------------------------- | ---------------------------------------- | -------- | ------------------------------------------------------- |
| [rsbuild-plugin-istanbul](https://github.com/canyon-project/canyon/blob/main/istanbul/rsbuild-plugin-istanbul/src/index.ts)                     | rsbuild Babel plugin instrumentation     | Failed   | Rspack transforms code earlier; source maps unavailable |
| Manual Babel transform in rspack              | Babel transform applied in rspack plugin | Working  | Functional but performance untested                     |
| SWC plugin [swc-plugin-coverage-instrument](https://github.com/kwonoj/swc-plugin-coverage-instrument) | Native SWC instrumentation               | Unusable | swc-core version mismatch with rspack                   |

Currently, the manual Babel transform in rspack is used.

---

## Future Plans

* Add V8 coverage support
* Support configurable reporters
* Benchmark instrumentation strategies

---

## Appendix: Implementation Notes

Originally tried `rsbuild-plugin-istanbul` but rspack’s pre-transform breaks source map chain. Switched to manual Babel transform inside rspack plugin. Also attempted SWC plugin but incompatible versions blocked usage.

travzhang:
That's amazing — just updated earlier this morning! I'll give it a try sometime this week.

[...]

PR #399, opened 2025-07-20
add istanbul coverage provider

travzhang:
## Related Links

https://github.com/web-infra-dev/rstest/issues/398

## Checklist


packages/core/package.json
```diff
@@ -60,8 +60,14 @@
...
     "birpc": "2.4.0",
     "chai": "^5.2.1",
+    "istanbul-lib-coverage": "^3.2.2",
+    "istanbul-lib-instrument": "^6.0.3",
+    "istanbul-lib-report": "^3.0.1",
+    "istanbul-lib-source-maps": "^5.0.6",
+    "istanbul-reports": "^3.1.7",
     "pathe": "^2.0.3",
     "std-env": "^3.9.0",
+    "swc-plugin-coverage-instrument": "^0.0.28",
```
9aoy: Builtin istanbul coverage directly into Rstest is too heavy. Rstest can provide a separate package for users to install as needed. When users set `provider: 'istanbul'`, rstest can import the relevant transform plugin and provider from `@rstest/coverage-istanbul`.

travzhang: 👌

packages/core/src/core/rsbuild.ts
```diff
@@ -221,6 +222,27 @@ export const prepareRsbuild = async (
                 ...(config.module.parser.javascript || {}),
               };
 
+              if (coverage?.enabled && coverage?.provider === 'istanbul') {
+                const swcPluginCoverageInstrument = `${process.cwd()}/node_modules/@rstest/coverage-istanbul/node_modules/swc-plugin-coverage-instrument/target/wasm32-wasip1/release/swc_plugin_coverage.wasm`;
```
9aoy: This path should be imported from `@rstest/coverage-istanbul` rather than hard-coded in rstset.

And `target/wasm32-wasip1/release/swc_plugin_coverage.wasm` is the default export of `swc-plugin-coverage-instrument`

image

packages/core/src/core/rsbuild.ts
```diff
@@ -221,6 +222,27 @@ export const prepareRsbuild = async (
...
+                config.module.rules ??= [];
+                config.module.rules.push({
+                  test: /\.(js|ts)$/,
+                  exclude: [/node_modules/],
```
9aoy: some test files should be excluded.

packages/core/src/coverage/index.ts
```diff
@@ -0,0 +1,23 @@
+import type { CoverageOptions } from '../types/coverage';
+import type { CoverageProvider } from './istanbul';
+import { IstanbulCoverageProvider } from './istanbul';
```
9aoy: `IstanbulCoverageProvider` should also extract to `@rstest/coverage-istanbul`

PR #405, opened 2025-07-21
feat: add `RSTEST` flag

9aoy:
## Summary

`process.env.RSTEST` can be used to determine if it is an rstest environment.

```ts
if (process.env.RSTEST) {
   // do something...
}
```
## Checklist


packages/core/src/core/rsbuild.ts
```diff
@@ -133,6 +133,7 @@ export const prepareRsbuild = async (
   const writeToDisk = dev.writeToDisk || debugMode;
 
   const rsbuildInstance = await createRsbuild({
+    callerName: 'rstest',
```
chenjiahan: should be split into a separate PR

PR #446, opened 2025-08-05
docs: add note for `describe.skip`

9aoy:
## Summary

The `skip` tag is only used to skip test cases, and the code inside the describe block will still be executed.

```ts
describe.skip('a', () => {
  console.log('will run');
  test('b', () => {
    console.log('will not run');
    expect(0).toBe(0);
  });
});
```

## Checklist


njzydark:
If I want to prevent test code from executing, are there related APIs that can achieve this? Like .todo?

9aoy:
> If I want to prevent test code from executing, are there related APIs that can achieve this? Like .todo?

None at the moment.

PR #475, opened 2025-08-14
feat: re-run on demand in watch mode.

fi3ework:
## Summary

Rerun test files on demand, with following major modification:

1. using a new Rstest class to hold the persistent test results, which will be used to do the partial update with.
2. calculate the affected test entry files by the diff of the chunks, and rerun them as on-demand.

## Checklist


e2e/watch/index.test.ts
```diff
@@ -49,13 +53,15 @@ describe('watch', () => {
...
-    expect(cli.stdout).toMatch('Test Files 1 failed | 1 passed');
+    expect(cli.stdout).toMatch('Test Files 1 failed');
     expect(cli.stdout).toMatch('✗ bar > bar should be to bar');
+    expect(cli.stdout).toMatch(/Test files to re-run:\n.*bar\.test\.ts\n\n/);
 
     // delete
     cli.resetStd();
     fs.delete('./fixtures-test-0/bar.test.ts');
     await cli.waitForStdout('Duration');
+    expect(cli.stdout).toMatch('No test files are re-run.');
```
9aoy, line 64: can we log removed files?

e2e/watch/shortcuts.test.ts
```diff
@@ -78,15 +81,15 @@ describe('CLI shortcuts', () => {
...
     await cli.waitForStdout('press h to show help');
-
+    expect(cli.stdout).toMatch('Fully run test files for first run.');
     cli.resetStd();
 
     // rerun failed tests
     cli.exec.process!.stdin!.write('f');
     await cli.waitForStdout('Duration');
     expect(cli.stdout).toMatch('Tests 1 failed');
+    expect(cli.stdout).toMatch('Run all tests.');
```
9aoy: This log is not accurate because only failed tests are rerun

packages/core/src/core/runTests.ts
```diff
@@ -109,18 +116,43 @@ export async function runTests(context: RstestContext): Promise<void> {
...
       hash,
+      isFirstRun,
+      affectedEntries,
+      deletedEntries,
     } = await getRsbuildStats({ fileFilters });
     const testStart = Date.now();
 
+    let finalEntries: EntryInfo[] = entries;
+    if (mode === 'on-demand') {
+      if (isFirstRun) {
```
9aoy, line 133: this flag seems to be available via `isFirstCompile`

e2e/watch/index.test.ts
fi3ework: it only will log when DEBUG is enabled, i think it's okay to preserve for test and debug.

e2e/watch/shortcuts.test.ts
fi3ework: changed based on `fileFilters`

9aoy (approved):

PR #504, opened 2025-08-22
fix: use clearScreen instead of clearLogs

9aoy:
## Summary

use clearScreen instead of clearLogs.

## Checklist


fi3ework (commented):
is the `console.clear` in `packages/core/src/core/cliShortcuts.ts` should be repalced as well?

9aoy:
> is the `console.clear` in `packages/core/src/core/cliShortcuts.ts` should be repalced as well?

nice catch!

PR #521, opened 2025-08-28
enhance: filter file based on workspace's root

9aoy:
## Summary

Filter file based on workspace's root. 

Previously, when rstest matched test files in projects, it converted the project root to a relative path. Therefore, if the project root was `project/core`, `npx rstest core` would not match all test files under `project/core`. 
This now uses the workspace root as a unified root path for matching.

before:

image

after:
image

## Checklist


fi3ework (commented):
should we update the match rule on documentation alongside?

PR #618, opened 2025-10-15
feat: drop support of async mock factory

fi3ework:
## Summary

implementing async mock factories comes at a significant cost. besides requiring extensive runtime modifications, properly handling circular dependencies necessitates deep AST analysis and transformations. this would severely impact performance, and because static imports are converted to dynamic imports, the test files' build would inevitably diverge from the source code, leading to unknown errors.

we are dropping support for async mock factories. after investigation, we found that async mock factories are primarily used to configure partial mocks using `importActual`, and the asynchronous nature of `importActual` necessitates an async mock factory.

in this PR, we introduce a new synchronous replacement for `rs.importActual` by using import attributes:

```ts
import * as foo from 'foo' with { "rstest": "importActual" }
rs.mock('foo', ()=> {
  return { ...foo, extra: 1 }
})
```

and, of course, `rs.importActual` will be retained. You can continue to use it within `it / test` test functions to import original modules that are not being mocked. however, it can no longer be used within mock factories, as the second parameter of the mock factory no longer accepts async functions.

## Checklist


e2e/mock/tests/doMockRequire.test.ts
```diff
@@ -1,6 +1,6 @@
 import { expect, rs, test } from '@rstest/core';
 
-test('doMockRequire works', () => {
```
9aoy, line 3: This logic doesn't seem asynchronous, so why does it need `async`?

website/docs/zh/config/test/importActualMethods.mdx
```diff
@@ -0,0 +1,46 @@
+---
+overviewHeaders: []
+---
+
+# importActualMethods
+
+- **类型：** `('importAttributes' | 'query')[]`
+- **默认值：** `['importAttributes', 'query']`
```
9aoy: If `['importAttributes', 'query']` is the default value, I don't think there is any need to add a configuration here.

e2e/basic/test/index.test.ts
```diff
@@ -24,8 +24,8 @@ describe('Index', () => {
   });
 
   it('should use require.resolve correctly', async () => {
-    expect(
-      require.resolve('../src/index.ts').endsWith('index.ts'),
-    ).toBeTruthy();
+    const resolved = require.resolve('../src/index.ts');
+    // TODO: can't write as  require.resolve('../src/index.ts').endsWith('index.ts')
+    expect(resolved.endsWith('index.ts')).toBeTruthy();
```
9aoy, line 29: what happens with `require.resolve().endsWith` ? 🥹

fi3ework: this is not related to mock logic, it will break even only bumping Rsbuild. should be related to a Rspack parser regression.

website/docs/zh/config/test/importActualMethods.mdx
fi3ework: for now, only to make `vue-loader` pass, as it hard coded allowed properties which doesn't include `with`. https://github.com/web-infra-dev/rstest/pull/618/files#diff-2f01f71a910f9b48ee94bd055ad54b6855171c7327268ef2f0ab563d4d15aa18R7

packages/core/src/core/rsbuild.ts
```diff
@@ -279,6 +283,21 @@ export const createRsbuildServer = async ({
...
+          .plugin('RemoveDuplicateModulesPlugin')
+          .use(rspack.experiments.RemoveDuplicateModulesPlugin);
+
+        // add mock-loader to this rule
+        chain.module
+          .rule('rstest-mock-module-doppelgangers')
+          .test(/\.(?:js|jsx|mjs|cjs|ts|tsx|mts|cts)$/)
+          .with({ rstest: 'importActual' })
+          .use('mock-loader')
+          .loader(path.resolve(__dirname, './mockLoader.mjs'))
```
9aoy: According to the semantics of `importActual` , it should not be mocked. Wouldn't `unmock-loader` make more sense?

fi3ework: what about `import-actual-loader`, it's a dedicated loader for `importActual` and won't be used elsewhere.

9aoy (approved):

PR #621, opened 2025-10-16
fix: out-of-memory in large project

9aoy:
## Summary

Fix `out-of-memory` error in large project. Get test dist assets on demand from memoryFs instead of getting all assets at once.
- Converted asset/sourcemap access from synchronous to async on-demand fetching
- Added memory monitoring utility (isMemorySufficient) to detect memory constraints
- Add new RPC method `getAssetsByEntry` to fetch assets in worker when memory is insufficient

img_v3_02pr_08513ea0-b083-46a7-ad2e-7c4605755d2g

This PR will cause a small performance drop, but will make rstest more usable in large projects.
rsbuild benchmark: 
image

## Checklist

### DIFF
--- a/packages/core/src/core/rsbuild.ts
+++ b/packages/core/src/core/rsbuild.ts
@@ -475,7 +475,7 @@ export const createRsbuildServer = async ({
         content = sourceMap;
       }
 
-      enableAssetsCache && cachedSourceMaps.set(name, content!);
+      enableAssetsCache && content && cachedSourceMaps.set(name, content);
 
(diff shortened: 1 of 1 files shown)


fi3ework (commented):
is the decline in performance due to RPC communication?

9aoy:
> is the decline in performance due to RPC communication?

A little decline comes from memory monitoring, and more degradation comes from RPC communication, but it does not happen in small projects and is better than OOM.

PR #685, opened 2025-11-11
feat(vscode): remove unnecessary logger code

claneo:
## Summary

1. Using VSCode builtin log output instead
2. Only external resources, like temporary files, require cleanup

image

## Checklist


fi3ework (approved):
wow, didn't know the built-in log level before, thank you.

PR #691, opened 2025-11-12
feat(vscode): communicate with worker using ipc instead of websocket

claneo:
## Summary

1. Replace websocket with ipc, removed unnessary dependencies
2. A RPC style api
3. Reusable logger for both masetr and worker

## Checklist

### DIFF
--- a/packages/vscode/src/worker/index.ts
+++ b/packages/vscode/src/worker/index.ts
@@ -1,6 +1,11 @@
 import { pathToFileURL } from 'node:url';
-import { WebSocket } from 'ws';
-import type { WorkerInitData, WorkerRunTestData } from '../types';
+import { createBirpc } from 'birpc';
+import type { RstestApi } from '../master';
+import type {
+  WorkerEventFinish,
+  WorkerInitData,
+  WorkerRunTestData,
+} from '../types';
 import { logger } from './logger';
 import { VscodeReporter } from './reporter';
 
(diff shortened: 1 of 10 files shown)


packages/vscode/src/shared/rpc.ts
```diff
@@ -0,0 +1,108 @@
+import type { ChildProcess } from 'node:child_process';
+
+export function expose(
```
fi3ework: i think it's better to use https://github.com/antfu-collective/birpc instead of maintaining the RPC methods.

claneo: Good idea. I only knew about comlink before, which can only be used in workers.

fi3ework: birpc is also being used in `@rstest/core` now. 😃

packages/vscode/src/worker/index.ts
```diff
@@ -1,33 +1,25 @@
...
-        const data: WorkerRunTestData = _data;
-        this.runTest(data);
-      }
-    });
-  }
-
   public async runTest(data: WorkerRunTestData) {
     logger.debug('Received runTest request', JSON.stringify(data, null, 2));
+    const { promise, resolve, reject } =
+      Promise.withResolvers<WorkerEventFinish>();
```
fi3ework: `Promise.withResolvers` is introduced in node22, and it runs in child process, the node spawn here https://github.com/web-infra-dev/rstest/blob/main/packages/vscode/src/master.ts#L188 is determined by user's node version. 

i think we need to VS Code's bundled node bin path, `process.execPath`.

claneo: To use the node bundled with VSCode, just change `spawn` to `fork`.

However, I think it's better to use the user's node version, allowing users to control the version themselves. I'll remove the use of `Promise.withResolvers`.

fi3ework (approved):

fi3ework:
thank you for your contribution. 😇

fi3ework:
there's a conflict, after resolving, we're good to merge.

claneo:
done

fi3ework:
thank you for the work again.

PR #696, opened 2025-11-13
chore: update vscode extension build script

9aoy:
## Summary

update vscode extension build script, fix some build warnings.

image

## Checklist

### DIFF
--- a/packages/vscode/rslib.config.mts
+++ b/packages/vscode/rslib.config.mts
@@ -50,7 +50,7 @@ export default defineConfig({
   ],
   tools: {
     rspack: {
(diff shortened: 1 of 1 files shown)


packages/vscode/rslib.config.mts
fi3ework, line 1: `mts` is good, but just curious, why it's renamed to `mts`.

9aoy: To fix this warning:
image

fi3ework: got it. the package is not declared as `"module"`

PR #704, opened 2025-11-19
fix: suppress ESModulesLinkingError for exports that might be implemented in mock

9aoy:
## Summary

suppress ESModulesLinkingError for exports that might be implemented in mock.

```bash
ESModulesLinkingError: export 'mockApiClient' (imported as 'mockApiFile') was not found in '../../services' (possible exports...
```

mock modules:

```ts
class MockApiClient extends ApiClient {
  fetch = rstest.fn().mockResolvedValue({ json: () => Promise.resolve({ data: [] }) })
}
export const mockApiClient = new MockApiClient({ baseDomain: 'https://example.com' });
```

## Related Links

https://github.com/web-infra-dev/rstest/discussions/698

## Checklist

### DIFF
--- a/e2e/mock/fixtures/esModulesLinkingError/index.test.ts
+++ b/e2e/mock/fixtures/esModulesLinkingError/index.test.ts
@@ -7,7 +7,7 @@ rs.mock('./services', () => ({
 }));
 
 describe('Mock Module EsModulesLinkingError', () => {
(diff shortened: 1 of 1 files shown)


chenjiahan (commented):
Should we use `fix` instead of `chore` in the PR title?

PR #716, opened 2025-11-26
chore: support `--bail` without value

9aoy:
## Summary

- support `--bail` without value (default `1`)
- add test bail tip.

image

## Checklist

### DIFF
--- a/e2e/bail/index.test.ts
+++ b/e2e/bail/index.test.ts
@@ -8,7 +8,7 @@ const __dirname = dirname(__filename);
 
 describe('test bail option', () => {
   it('should not run all tests when `--bail [number]`', async () => {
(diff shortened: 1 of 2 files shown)


e2e/bail/index.test.ts
```diff
@@ -7,7 +7,7 @@ const __filename = fileURLToPath(import.meta.url);
 const __dirname = dirname(__filename);
 
 describe('test bail option', () => {
-  it('should not run all tests when bail option is set', async () => {
+  it('should not run all tests when `--bail <number>`', async () => {
```
fi3ework:
```suggestion
  it('should not run all tests when `--bail [number]`', async () => {
```

PR #720, opened 2025-11-27
feat(vscode): report test run progress

claneo:
## Changes Summary

1. core
   1. Support pass logger to default reporter
   2. Add two new reporter hooks: `onTestSuiteStart` and `onTestSuiteResult`
   3. Support match test case with `testNamePattern` exactly (This is still a temporary solution that cannot handle cases where the name contains `>`. I plan to support matching by arrays instead of strings in the future.)
   4. Stop stringify `actual` and `expected` value in `error` object
1. vscode extension
   1. New features, see below
   2. Simplified `updateFromContents`, now ancestors stack is poped when exit astnode
   3. Enqueued all test items before run them

## New Features

1. Realtime test result
6. Pipe default reporter to test results panel
7. Place error message at right location
8. Display error stack if exist
9. Display expected and actual value if it is a mismatch error
10. Support update snapshot if it is a snapshot mismatch error

https://github.com/user-attachments/assets/464e4fdd-c18e-42fc-a878-da5d0663952f

## Problems

1. Sourcemap does not work if debugger attached, causes error stack is wrong
2. Is it ok to import core source code from vscode extension directly?

## Checklist

### DIFF
--- a/packages/core/src/runtime/runner/runner.ts
+++ b/packages/core/src/runtime/runner/runner.ts
@@ -359,7 +359,7 @@ export class TestRunner {
             result.errors?.push(...formatTestError(error));
           }
         }
-        result.duration = Date.now() - start;
+        result.duration = RealDate.now() - start;
         result.status = result.errors?.length
           ? 'fail'
           : getTestStatus(results, defaultStatus);


packages/core/src/types/config.ts
```diff
@@ -189,6 +189,10 @@ export interface RstestConfig {
    * Run only tests with a name that matches the regex.
    */
   testNamePattern?: string | RegExp;
+  /**
+   * Match `testNamePattern` exactly, test suite and case names should be joined by ' > ' (with spaces)
```
fi3ework: `exact` exists solely to serve `testNamePattern`. i think `exact` could be tried within `testNamePattern` (e.g., `{ testNamePattern: { matcher: string | RegExp; exact? boolean}`). 

however, i wonder if RegExp can currently achieve the same purpose (using `$` and `^`), so we wouldn't need to add a new configuration option.

claneo: > however, i wonder if RegExp can currently achieve the same purpose (using $ and ^), so we wouldn't need to add a new configuration option.

That sounds feasible, although it's still not accurate, but it's not much different from introducing the `exact` option.

fi3ework: if the `exact` currently is mainly for vscode extension, i think we could resue the RegExp. we could introduce `exact` field if meets real user requirements.

packages/core/src/runtime/runner/runner.ts
```diff
@@ -338,12 +356,18 @@ export class TestRunner {
             }
           } catch (error) {
             // AfterAll failed does not affect test case results
-            errors.push(...formatTestError(error));
+            result.errors?.push(...formatTestError(error));
           }
         }
+        result.duration = Date.now() - start;
```
fi3ework: should use `RealDate`, just like `start`.

9aoy:
👍

PR #738, opened 2025-12-05
feat(vscode): support run tests with debug or coverage, list test items in tree structure

claneo:
depends on #720 

## New Features

1. Better test list view  
    1. Display test items in a tree structure instead of the previous flat list  
    2. Hide the workspace node if there's only one workspace open  
    3. Hide the project node if there's only one project and it is placed at the root of the workspace  
2. Spawn a separate worker every time the test is run  
    1. Support debugging tests by starting the worker with the `--inspect-wait` argument  
    2. Support test run cancellation by killing the worker  
3. Support running tests with coverage  
4. Use the capabilities of rstest itself to execute multiple test files in parallel  
5. The test files scan result is exactly the same as rstest. Now the extension reads include/exclude from the config file, scans with `tinyglobby`, then watches changes with the VSCode watcher, and filters files with `picomatch`  

https://github.com/user-attachments/assets/396cf218-ddd6-454c-8f4d-cd79cff39f97

## Checklist

### DIFF
--- a/packages/vscode/src/project.ts
+++ b/packages/vscode/src/project.ts
@@ -205,8 +205,12 @@ export class Project implements vscode.Disposable {
     return this.testItem?.children || this.parentCollection;
   }
   private async startWatchingWorkspace(root: vscode.Uri) {
-    const isInclude = picomatch(this.include, { cwd: root.fsPath });
-    const isExclude = picomatch(this.exclude, { cwd: root.fsPath });
+    const matchInclude = picomatch(this.include);
+    const matchExclude = picomatch(this.exclude);
+    const isInclude = (uri: vscode.Uri) => {
+      const relativePath = path.relative(root.fsPath, uri.fsPath);
+      return matchInclude(relativePath) && !matchExclude(relativePath);
+    };
 
     const files = await glob(this.include, {
       cwd: root.fsPath,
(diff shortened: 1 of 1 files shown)


packages/vscode/src/project.ts
```diff
@@ -100,130 +124,224 @@ export class WorkspaceManager implements vscode.Disposable {
...
-      async (globs, token) => {
-        const patterns = globs.map(
-          (glob) => new vscode.RelativePattern(this.root, glob),
-        );
+  get collection() {
+    return this.testItem?.children || this.parentCollection;
+  }
+  private async startWatchingWorkspace(root: vscode.Uri) {
+    const isInclude = picomatch(this.include, { cwd: root.fsPath });
+    const isExclude = picomatch(this.exclude, { cwd: root.fsPath });
```
fi3ework: picomatch doesn't respect https://github.com/micromatch/picomatch/issues/105 actually, so should pass a relative glob to `isInclude` instead of abs path.

fi3ework (approved):
LGTM, great work, and a huge thanks to @claneo.

fi3ework:
cc @9aoy , coverage reporter now can accept an instance, i think it's nice to have that. PTAL.

PR #739, opened 2025-12-06
feat: collect test location

claneo:
## Summary

1. Collect test location via error stack
2. `list` command
   1. Added new option `--printLocation`
   2. List test suites too
3. Added new reporter hook `onTestFileReady`, this will be used in vscode extension later

The location maybe inaccurate because rspack transpiled `describe()` call to `(0,rstest_core.describe)()`, so the location got from error stack is the end of the comma expression. Chained method is not affected expect `describe.for` which is tranpiled to `describe["for"]`.

```shell
$ cd e2e && npx rstest list ./build --printLocation
build/index.test.ts:9:9 > test build config
build/index.test.ts:10:20 > test build config > define config should work correctly
build/index.test.ts:10:20 > test build config > alias config should work correctly
build/index.test.ts:10:20 > test build config > plugin config should work correctly
build/index.test.ts:10:20 > test build config > tools/rspack config should work correctly
build/index.test.ts:10:20 > test build config > decorators config should work correctly
build/resolve.test.ts:9:3 > should resolve correctly 
build/runtimeImport/index.test.ts:8:3 > should load runtime deps correctly
build/runtimeImport/index.test.ts:13:3 > should load compile-time deps correctly
```

## Checklist

### DIFF
--- a/website/docs/en/guide/basic/cli.mdx
+++ b/website/docs/en/guide/basic/cli.mdx
@@ -102,6 +102,32 @@ $ npx rstest list --json
 $ npx rstest list --json=./output.json
 ```
 
+You can use `--includeSuites` to print test suites along side test cases:
+
+```bash
+$ npx rstest list
+
+# the output is shown below:
+a.test.ts > test a
+a.test.ts > test a > test a-1
+a.test.ts > test a-2
+b.test.ts > test b
+b.test.ts > test b > test b-1
+b.test.ts > test b-2
+```
+
+You can use `--printLocation` to print location of tests:
+
+```bash
+$ npx rstest list
+
+# the output is shown below:
+a.test.ts:4:5 > test a > test a-1
+a.test.ts:9:3 > test a-2
+b.test.ts:4:5 > test b > test b-1
+b.test.ts:9:3 > test b-2
+```
+
 ## CLI options
 
 Rstest CLI provides several common options that can be used with all commands:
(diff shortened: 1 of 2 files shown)


packages/core/src/runtime/runner/runtime.ts
```diff
@@ -475,13 +489,34 @@ export const createRuntimeAPI = ({
     runtimeConfig,
   });
 
+  // TODO allow disable get location via config
+  const getLocation = (): Location | undefined => {
```
9aoy, line 486: `location` should disable by default since it's not a common use case, and we can add a config option to show test location via config, like vitest's [includeTaskLocation](https://vitest.dev/config/includetasklocation.html#includetasklocation).

e2e/list/index.test.ts
```diff
@@ -73,8 +84,14 @@ describe('test list command', () => {
 
     expect(logs).toMatchInlineSnapshot(`
       [
+        "a.test.ts > test a",
```
9aoy: test suites should not be listed to avoid too much verbosity, we can add a cli option to include suites, eg `--includeSuite`

claneo: I was misled by vitest. The example in its documentation outputs suites, so I was planning to align its behavior. But I just tested it, and it doesn't actually output suites.

claneo:
> It would be even better if there were some documentation available~

I'll add documentation for `printLocation` and `includeSuites`. As for `includeTaskLocation`, no reporter currently uses the `location` field, so I think we can add documentation for it when a reporter needs it in the future, like the HTML reporter.

9aoy (approved):

PR #754, opened 2025-12-10
feat(vscode): support collect tests from runtime

claneo:
## Summary

depends on #738, #739

1. Added a new configration to control how to collect test cases, ast or runtime
2. Test cases will always be updated from runtime when running test file

All test cases in this repo can run in vsocde extension now.

image

## Checklist

### DIFF
--- a/packages/vscode/src/project.ts
+++ b/packages/vscode/src/project.ts
@@ -266,9 +266,7 @@ export class Project implements vscode.Disposable {
         const watcher = vscode.workspace.createFileSystemWatcher(
           new vscode.RelativePattern(root, '**'),
         );
-        this.cancellationSource.token.onCancellationRequested(() =>
-          watcher.dispose(),
-        );
+        token.onCancellationRequested(() => watcher.dispose());
 
         // TODO delay and batch run multiple files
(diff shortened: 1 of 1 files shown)


packages/vscode/src/project.ts
```diff
@@ -212,68 +213,118 @@ export class Project implements vscode.Disposable {
...
+
+        // start watching test file change
+        // while createFileSystemWatcher don't support same glob syntax with tinyglobby
+        // we can watch all files and filter with picomatch later
+        const watcher = vscode.workspace.createFileSystemWatcher(
+          new vscode.RelativePattern(root, '**'),
+        );
+        this.cancellationSource.token.onCancellationRequested(() =>
+          watcher.dispose(),
+        );
```
fi3ework: The `watcher` should be disposed when the config changes, not just when the Project is disposed. Use `token.onCancellationRequested` instead of `this.cancellationSource.token.onCancellationRequested`, otherwise switching between `ast` and `runtime` modes will accumulate multiple active watchers.

```js
token.onCancellationRequested(() => watcher.dispose());
```

PR #821, opened 2025-12-31
feat(coverage-istanbul): transform coverage using source map

claneo:
## Summary

This PR extracts `sourceMapUrl` from js files and uses `istanbul-lib-source-maps` to map coverage back to the original source code.

## Checklist


9aoy:
👍

PR #842, opened 2026-01-08
feat(core): first-class Module Federation support

ScriptedAlchemy:
This PR adds first-class Module Federation (MF) support/compatibility to Rstest, with a new `federation` config option and a full set of examples/tests.

### Config + validation
- Add `federation?: boolean` config option (default `false`).
- When `federation: true` and `output.module` is not explicitly set, default to CommonJS (`output.module = false`).
- Fail fast if federation is enabled while ESM output is enabled.

### Rsbuild/Rspack federation compatibility
- Add Rsbuild plugin `rstest:federation-compat` to improve MF Node runtime compatibility:
  - set async-node compilation target
  - disable `splitChunks` for more reliable MF runtime loading
  - validate `ModuleFederationPlugin` config for DOM environments

### Federation-aware externals
- Keep MF runtime packages bundled (e.g. `@module-federation/*`) when federation is enabled.
- Keep MF “loader-style” requests bundled (e.g. `...!=!data:text/javascript,...`).
- Keep unresolved remote-like specifiers bundled in federation mode so the runtime can resolve them.

### Runtime shims / environment gating
- Provide a safe `globalThis.__rstest_dynamic_import__` fallback for chunks evaluated via `vm/eval`.
- Add chunk-handler guards so MF bootstrap placeholders don’t break chunk loading before the runtime patches them.
- Propagate federation flag into the worker/global-setup worker to gate federation-only behavior.

### Federation without `writeToDisk`
- Stop forcing `dev.writeToDisk = true` solely due to federation; keep in-memory output by default.
- Add a worker-level virtual FS bridge to serve compiled assets from the in-memory `assetFiles` map when MF runtimes try to read chunks via Node `fs` APIs.
  - Only installed/active when `federation: true`.

### DOM error handling stability (tests + runtime)
- Stabilize default DOM error forwarding so `window.addEventListener('error', ...)` handlers reliably prevent “unhandled” errors in both jsdom and happy-dom.
- Avoid unnecessary global/window binding differences unless federation mode is enabled.

### Snapshot stability
- Normalize pnpm store paths in snapshots to avoid machine-specific snapshot diffs.

## `@rstest/browser`
- Pass `federation` through the browser host runtime config so the same federation-only behavior can apply in browser mode.

## Examples
- Add `examples/federation/*` (main-app + component-app + node-local-remote) demonstrating MF usage and providing coverage for SSR + dynamic/static remote loading.

## Verification
- `pnpm build`
- `pnpm test`
- `pnpm e2e`


packages/core/src/core/plugins/external.ts
```diff
@@ -106,18 +125,48 @@ export const pluginExternal: (context: RstestContext) => RsbuildPlugin = (
...
+                        ? data
+                        : data && typeof data.request === 'string'
+                          ? data.request
+                          : undefined;
+
+                    if (
+                      typeof req === 'string' &&
+                      shouldKeepBundledForFederation(req)
+                    ) {
+                      return callback();
```
9aoy: Returning `callback(undefined, false)` from an external function tells Rspack to skip the remaining external configuration and bundle the dependency instead.

With this behavior, the federation-related external checks can be moved into the `federationPlugin`, rather than being coupled to the externals plugin, and without proxy the user’s external function configuration.

ScriptedAlchemy:
Closing in favor of #1407, a cleaned-up rewrite of this Module Federation support (same feature surface: `federation` config + `--federation` flag, runtime shims, examples, e2e, docs). The branch is kept for reference.

PR #883, opened 2026-01-20
feat: unify browser mode and node mode summary report

9aoy:
## Summary

- unify browser mode and node mode summary report
- make browser mode and node mode parallel execution

before:
image

after:
image

## Checklist


packages/browser/src/hostController.ts
```diff
@@ -1531,7 +1568,7 @@ export const runBrowserController = async (context: Rstest): Promise<void> => {
       waitUntil: 'load',
     });
 
-    logger.log(
+    logger.debug(
```
fi3ework: should not be modified

image

9aoy:
```suggestion
    logger.log(
```

PR #891, opened 2026-01-21
fix(core): lazy initialize StatusRenderer to prevent clearing early error messages

fi3ework:
## Summary

- Lazy initialize `StatusRenderer` in `DefaultReporter` to avoid early stdout/stderr interception
- Fix visibility of fatal errors (e.g., missing Playwright browsers) in interactive terminals

human input:

when playwright doesn't installed any browsers, the error message throw from `browser.launch` will be cleared by StatusRenderer, while no test file has started yet. so lazy initialize StatusRenderer to allow error message throw in Rsbuild initializing phase.

| Before | After |
|--------|-------|
|    image |     image  |

## Checklist

### DIFF
--- a/packages/core/src/reporter/index.ts
+++ b/packages/core/src/reporter/index.ts
@@ -39,16 +39,30 @@ export class DefaultReporter implements Reporter {
     this.config = config;
     this.options = options;
     this.testState = testState;
-    if (isTTY() || options.logger) {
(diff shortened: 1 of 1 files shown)


claneo:
The actual reason Playwright’s error messages are not shown is that the reporter’s `onTestRunEnd` is never invoked.

Lazy initialization of `StatusRenderer` can leave the console blank while a slow build step is running.

PR #916, opened 2026-01-28
feat: support filter test by full test name

9aoy:
## Summary

support filter test by full test name.

image

## Related Links

https://github.com/web-infra-dev/rstest/pull/911

## Checklist

- [ ] Tests updated (or not required).
- [ ] Documentation updated (or not required).

PR #926, opened 2026-01-30
fix(vscode): stderr logs not written to output

claneo:
## Summary

Added the`mergeStderrToStdout` option to merge stderr output into stdout, addressing the limitation that VSCode’s`appendOutput` supports only a single output stream.

This also fixes the issue where running a single test was being skipped, caused by #916, which changed the delimiter from two spaces to one space.

## Checklist

### DIFF
--- a/packages/vscode/src/worker/reporter.ts
+++ b/packages/vscode/src/worker/reporter.ts
@@ -22,15 +22,14 @@ export class ProgressReporter implements Reporter {
 }
 
 export class ProgressLogger {
-  stream = new Writable({
+  outputStream = new Writable({
     decodeStrings: false,
     write: (chunk, _encoding, cb) => {
       masterApi.onOutput.asEvent(chunk);
       cb(null);
     },
   });
-  outputStream = this.stream;
-  errorStream = this.stream;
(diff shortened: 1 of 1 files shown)


packages/core/src/reporter/windowedRenderer.ts
```diff
@@ -28,6 +28,11 @@ export interface Options {
   logger: {
     outputStream: Writable;
     errorStream: Writable;
+    /**
+     * When enabled, all intercepted `process.stderr` output will be forwarded
+     * to `outputStream` (stdout) instead of `errorStream` (stderr).
+     */
+    mergeStderrToStdout?: boolean;
```
9aoy: This parameter looks a bit odd. Is this a standard implementation? Can this feature be achieved by passing the same stream param to both outputStream and errorStream (errorStream: outputStream)?

claneo: Yes you're right. I misread the `WindowRenderer` implementation and assumed that `stdout` and `stderr` would be counted toward `windowHeight`.

9aoy:
👍

Issue #745, opened 2025-12-08
[Bug]: `rs.mock()` loads real implementation (executing side effects)


jhage-snapsheet:
### Version

```sh
System:
    OS: macOS 15.6
    CPU: (14) arm64 Apple M4 Pro
    Memory: 92.91 MB / 24.00 GB
    Shell: 5.9 - /bin/zsh
  Browsers:
    Chrome: 142.0.7444.177
    Firefox: 141.0
    Safari: 18.6
  npmPackages:
    @rstest/core: latest => 0.7.1
```

### Details

As far as I can tell, the rstest implementation of mocking is actually loading the "source" file before applying the mock. I don't see that this is happening with jest

In our case we have endless side effects in our "source" files and can't justify refactoring right now

### Reproduce link

https://github.com/jhage-snapsheet/rstest-side-effect-mock

### Reproduce Steps

1. `npm run test` or `DEBUG=rstest npm run test`
2. observe "touches" variables are incremented instead of undefined
3. `npm run test:jest`
4. observe "touches" variables are not incremented

YanxinTang:
Can we provide an option in rs.mock to disable importing the real module？@9aoy

9aoy:
> my repro still fails with latest rstest, is there a bug in my repro?

<img width="772" height="71" alt="Image" src="https://github.com/user-attachments/assets/e38cb434-e08a-4be4-9fdf-d101560b5083" />

@jhage-snapsheet You should replace `jest.mock` with `rs.mock`, as the mocking mechanism relies on AST transformation to work correctly.

jhage-snapsheet:
Just checking in @9aoy , sorry!

PR #934, opened 2026-02-03
fix: should not load real implementation when module is mocked

9aoy:
## Summary

should not load real implementation when module is mocked.
- Modified `rs.mock()` to check the module cache first and avoid loading modules unless necessary (e.g., for spy/mock options)
- Updated module factory handling to set factories instead of directly modifying cache, ensuring consistent behavior
- Enhanced webpack runtime protection to prevent mocked module factories from being overwritten during chunk loading

## Related Links

fix: https://github.com/web-infra-dev/rstest/issues/745

## Checklist


fi3ework:
👍🏻

PR #949, opened 2026-02-06
feat: support capture stderr when 'Worker exited unexpectedly' error

9aoy:
## Summary

Captures related stderr output from child processes so that it can be attached to "Worker exited unexpectedly" errors.

image

## Checklist


packages/core/src/pool/stderrCapture.ts
fi3ework, line 1: this looks too complex 🤔

9aoy: tinypool pipes the child process stderr directly to the parent process, so we need some extra handling here. 😢

Issue #649, opened 2025-10-27
[Feature]: Mock all files with certain extension


akcyp:
### What problem does this feature solve?

In Jest it's common to create following configuration to ignore / mock files with certain file extensions (css, less, scss, png, etc.).
```json
{
  "moduleNameMapper": {
    "\\.(css|less|scss)$": "<rootDir>/src/Mocks/styleMock.js" // module.exports = {}
  }
}
```
It seems that in the case of rstest this is not so easy to achieve without a custom plugin.

### What does the proposed API look like?

Support regexpes in `resolve.alias` or create native "ignore" plugin that will override files content with empty export.

9aoy:
This can be achieved using rspack [NormalModuleReplacementPlugin](https://rspack.rs/plugins/webpack/normal-module-replacement-plugin).
```ts
export default defineConfig({
  tools: {
    rspack: (config, { rspack}) => {
      config.plugins.push(
       new rspack.NormalModuleReplacementPlugin(/\.(css|less|scss)/, require.resolve('./src/Mocks/styleMock.js'))
      );
    }
  }
});
```

akcyp:
Thanks for the reply.
I didn't know about this plugin, maybe it would be worth considering adding this information to the migration guide from Jest?
I'm closing this thread.

fi3ework:
Rstest now handles CSS, CSS Modules, preprocessor styles and assets out of the box (#1825), so no stub is needed. To replace style imports by extension, see https://rstest.rs/guide/basic/css#replace-style-imports-by-extension. Regex `moduleNameMapper` is tracked in #990.

PR #990, opened 2026-02-27
feat: support `resolve.moduleNameMapper`

9aoy:
## Summary

Implement `resolve.moduleNameMapper` configuration option for Jest compatibility, allowing users to map module imports using **regex patterns**. This is similar to Jest's [moduleNameMapper](https://jestjs.io/docs/configuration#modulenamemapper-objectstring-string--arraystring) configuration.

```ts title="rstest.config.ts"

export default defineConfig({
  resolve: {
    moduleNameMapper: {
      // Map exact module name
      '^module-a$': '<rootDir>/mocks/module',
      // Map with capture groups
      '^@utils/(.*)$': '<rootDir>/src/utils/$1',
      // Stub CSS imports
      '\\.(css|less|scss)$': 'identity-obj-proxy',
    },
  },
});
```

## Related Links

close https://github.com/web-infra-dev/rstest/issues/649

## Checklist

### DIFF
--- a/packages/core/src/core/plugins/external.ts
+++ b/packages/core/src/core/plugins/external.ts
@@ -3,27 +3,8 @@ import type { RsbuildPlugin, Rspack } from '@rsbuild/core';
 import type { RstestContext } from '../../types';
 import { ADDITIONAL_NODE_BUILTINS, castArray } from '../../utils';
 
-/**
- * Check if a request matches any moduleNameMapper pattern.
- * If it does, don't externalize it - let NormalModuleReplacementPlugin handle it.
- */
-const matchesModuleNameMapper = (
-  request: string,
-  moduleNameMapper: Record<string, string | string[]> | undefined,
-): boolean => {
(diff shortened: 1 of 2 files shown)


fi3ework:
i feel like these two options are too similar for users, especially since they are almost identical when not using regular expressions. what if we provide a built-in plugin called moduleNameMapper? this way, we can avoid extending `resolve` while ensuring an almost identical Jest API.

fi3ework:
> > maybe rsbuild-plugin-jest-compatible for future Jest fields, or rsbuild-plugin-test-compatible for even Vitest compatible.
> 
> I prefer `rsbuild-plugin-module-name-mapper`. It accurately describes the capability, aligns with Rsbuild's plugin- naming convention, and keeps the scope honest. The regex-based path replacement is genuinely useful beyond testing, so a focused name avoids boxing the plugin into a "Jest compat" niche.

makes sense to me.

Issue #1045, opened 2026-03-12
[Feature]: Supports control whether to collect external file coverage


9aoy:
### What problem does this feature solve?

Rstest's current default behavior for coverage collection differs from both Jest and Vitest when it comes to files outside the project root (e.g., modules from sibling workspace packages in a monorepo).

**Current behavior comparison:**

| Framework | Default behavior | External coverage control |
|-----------|-----------------|--------------------------|
| **Jest** | External files excluded | No built-in option |
| **Vitest** | External files excluded | `coverage.allowExternal` (default `false`) |
| **Rstest (current)** | External files **included** | No option available |

This default-on behavior is inconsistent with the ecosystem convention (both Jest and Vitest default to excluding external files). It can also produce noisy reports that include coverage for `node_modules` symlinked packages or unrelated workspace dependencies that the user doesn't intend to track.

**Referenced from:** [#85 (comment)](https://github.com/web-infra-dev/rstest/issues/85#issuecomment-4041503743)

### What does the proposed API look like?

We should align with the Jest/Vitest default (exclude external files), and provide an option for users who need cross-package coverage in monorepo setups.

```ts
// rstest.config.ts
import { defineConfig } from '@rstest/core';

export default defineConfig({
  coverage: {
    enabled: true,
    // Allow coverage collection for files outside the project root
    allowExternal: true, // default: false
  },
});
```

PR #1050, opened 2026-03-12
feat: support `coverage.allowExternal` option

9aoy:
## Summary

Added `coverage.allowExternal` option to control whether coverage is collected for files outside the project root directory.

**Current behavior:** Rstest instruments all files processed during bundling, including files outside the project root (e.g., sibling workspace packages in a monorepo), resulting in noisy coverage reports.

**New behavior:** By default (`allowExternal: false`), files outside the project root are excluded from coverage reports, aligning with Jest and Vitest conventions. Users can opt in with `allowExternal: true` for monorepo cross-package coverage.

| `allowExternal` | `coverage.include` | External files in report? |
|:---:|:---:|:---:|
| `false` (default) | not set | No |
| `false` (default) | set | No |
| `true` | not set | Yes |
| `true` | set (matching external paths) | Yes (if matched) |

## Related Links

close https://github.com/web-infra-dev/rstest/issues/1045

## Checklist

### DIFF
--- a/website/docs/zh/config/test/coverage.mdx
+++ b/website/docs/zh/config/test/coverage.mdx
@@ -211,6 +211,8 @@ export default defineConfig({
 
 ### allowExternal
 
+<ApiMeta addedVersion="0.9.3" />
+
 - **类型：** `boolean`
 - **默认值：** `false`
 - **CLI：** `--coverage.allowExternal`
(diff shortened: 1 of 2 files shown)


website/docs/zh/config/test/coverage.mdx
```diff
@@ -208,6 +209,27 @@ export default defineConfig({
 });
 ```
 
+### allowExternal
```
fi3ework, line 212: can use `apiVersion` widget now, to demonstrate it's added since next release version.

PR #1072, opened 2026-03-19
feat(core): add output.bundleDependencies config option

### LINKED ISSUE #1064: [Bug]: jsdom breaks module mock
### Version

```sh
System:
    OS: macOS 15.7.2
    CPU: (14) arm64 Apple M4 Pro
    Memory: 136.38 MB / 24.00 GB
    Shell: 5.9 - /bin/zsh
  Browsers:
    Chrome: 145.0.7632.160
    Firefox: 147.0.1
    Safari: 26.1
  npmPackages:
    @rstest/adapter-rsbuild: ^0.2.2 => 0.2.2 
    @rstest/core: ^0.9.2 => 0.9.2
```

### Details

I tried to migrate from vite in my project, and found a strange issue with only one specific dependency `vue-router@4.5.1`. When I try to mock it, it seems that something is off with module resolution. I found that the bug appears only when I set testEnvironment to jsdom. Fixed by either disabling jsdom, OR changing the specifier on the mock decalration.

### Reproduce link

-

### Reproduce Steps

reproduces without a config file `rstest --include test.ts --testEnvironment jsdom`:

```js
import {rs} from '@rstest/core'
import * as vueRouterActual from 'vue-router' with { rstest: 'importActual' }
import {
    useRoute,
} from 'vue-router'


rs.mock('vue-router', () => {
    return {
        useRoute: rs.fn(),
    }
})

console.log(vueRouterActual.useRoute === useRoute) // true, but expected false
```

If I change the import specifier to `vue-router/dist/vue-router` it works!

```js
import {rs} from '@rstest/core'
import * as vueRouterActual from 'vue-router/dist/vue-router' with { rstest: 'importActual' }
import {
    useRoute,
} from 'vue-router'


rs.mock('vue-router/dist/vue-router', () => {
    return {
        useRoute: rs.fn(() => mockedRoute),
    }
})

console.log(vueRouterActual.useRoute === useRoute) // false as expected
```

9aoy:
### Background

Rstest ties the dependency bundling strategy to `testEnvironment` — browser-like environments (jsdom, happy-dom) bundle all `node_modules`, while Node externalizes them. Users have no straightforward way to override this without manually configuring `output.externals` per-package or reaching into `tools.rspack`. See #1064.

### Implementation

- Added `output.bundleDependencies` config option (`boolean`, 'true' for browser-like test environments, 'false' for node  test environment) to `ProjectConfig` type definition.
- Updated `pluginExternal` to read `bundleDependencies` and decide externalization independently of `testEnvironment.name`.
- Added E2E tests that verify build output: check that `strip-ansi` source code is inlined (bundled) or absent (externalized).
- Added documentation for both EN and ZH, including browser mode caveat and relationship with `output.externals`.

### User Impact

Users can now set `output.bundleDependencies: false` in jsdom to externalize all dependencies, or `true` in Node to bundle everything, decoupling the bundling strategy from `testEnvironment`.

## Related Links

- Closes #1064

## Checklist

### DIFF
--- a/website/docs/zh/config/build/output.mdx
+++ b/website/docs/zh/config/build/output.mdx
@@ -156,7 +156,7 @@ export default defineConfig({
 当未设置该选项时，rstest 会在类浏览器测试环境（jsdom、happy-dom 等）中打包依赖，在 `node` 环境中将其外部化。
 
 :::warning
-此选项仅适用于非浏览器模式。在[浏览器模式](/guide/browser/index)下，所有依赖始终会被打包，因此不支持 `output.bundleDependencies: false`。
+此选项仅适用于非浏览器模式。在[浏览器模式](/guide/browser-testing)下，所有依赖始终会被打包，因此不支持 `output.bundleDependencies: false`。
 :::
 
 此选项提供了一种简单的方式来覆盖与 `testEnvironment` 绑定的默认打包策略。例如，如果你使用 `jsdom` 但希望与 `node` 环境相同的外部化行为：
--- a/packages/core/src/core/plugins/external.ts
+++ b/packages/core/src/core/plugins/external.ts
@@ -112,11 +112,13 @@ export const pluginExternal: (context: RstestContext) => RsbuildPlugin = (
   setup: (api) => {
     api.modifyEnvironmentConfig((config, { mergeEnvironmentConfig, name }) => {
       const {
-        normalizedConfig: { testEnvironment, output: outputConfig },
+        normalizedConfig: {
+          testEnvironment,
+          output: { bundleDependencies } = {},
+        },
         outputModule,
       } = context.projects.find((p) => p.environmentName === name)!;
 
-      const bundleDependencies = outputConfig?.bundleDependencies;
       const shouldExternalize =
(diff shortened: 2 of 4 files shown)


website/docs/zh/config/build/output.mdx
```diff
@@ -143,6 +143,54 @@ export default defineConfig({
...
 ```
 
+## output.bundleDependencies
+
+- **类型：** `'auto' | boolean`
+- **默认值：** `'auto'`
+
+控制是否打包 `node_modules` 中的第三方依赖。
+
+- `'auto'`（默认）：在类浏览器测试环境（jsdom、happy-dom 等）中打包依赖，在 `node` 环境中将其外部化。
```
fi3ework: tbh, i don't think the code can run in browser mode if this field is set to `false` (no bundless mode for Rspack lol). so it seems that this option should always be `true` for browser mode.

so i'm wondering if it could be changed to the following style:

1. remove `auto`, only preserve the `boolean` type
2. declare this field not applicable for browser mode (there will be more other fields that are not supported, it's not the only one.）

PR #1074, opened 2026-03-19
perf(core): cache TraceMap and parsed sourcemaps, use Set for coverage file lookup

fi3ework:
## Summary

Add caching layers in the sourcemap resolution and error stack parsing paths
to eliminate redundant work when the same file is referenced by multiple
stack frames or errors within a single test run.

Not on the hot path today, but they are zero-cost correctness improvements:
- Function-scoped Map, created and discarded each call — no stale state, no watch-mode risk
- Eliminates redundant new TraceMap() (VLQ decode) and JSON.parse() when multiple stack frames reference the same file
- The effect is visible when a run produces many failures pointing to the same source files (e.g., a broken import causing 100+ test failures)
- Even when the savings are small, caching a pure lookup is the idiomatic pattern — it makes the intent clear and prevents future regressions if this path becomes hotter

Changes:

1. **Cache `TraceMap` per file** in both `parseErrorStacktrace` implementations
   (`packages/core/src/utils/error.ts` and `packages/core/src/reporter/md.ts`).
   Avoids redundant VLQ decoding when multiple stack frames originate from the
   same source file.

2. **Memoize `JSON.parse` in `getSourcemap`** (`packages/core/src/core/runTests.ts`).
   A per-run `parsedSourceMapCache` prevents repeated parsing of the same
   sourcemap JSON when the same file appears in multiple errors.

3. **Use `Set.has` instead of `Array.includes`** for coverage file lookup
   (`packages/core/src/coverage/generate.ts`). Converts O(n) linear scans to
   O(1) hash lookups when filtering covered/uncovered files.

4. **Fix `fullStack` condition** in `md.ts` reporter (`packages/core/src/reporter/md.ts`).
   Only re-parse the full stack when `error.fullStack` is truthy, avoiding
   unnecessary duplicate parsing when the full stack is not requested.

All caches are scoped to per-call or per-`run()` lifetime, so they are
automatically invalidated on each watch-mode rerun.

## Related Links

- Node perf optimization skill: `.agents/skills/node-perf-optimization/`

## Checklist

### DIFF
--- a/packages/core/src/core/runTests.ts
+++ b/packages/core/src/core/runTests.ts
@@ -426,30 +426,19 @@ export async function runTests(context: Rstest): Promise<void> {
         }
       }
 
-      // Memoize parsed sourcemaps to avoid redundant JSON.parse when the same
-      // file is referenced by multiple errors or stack frames.
-      const parsedSourceMapCache = new Map<string, SourceMapInput | null>();
-
       const getSourcemap = async (
         sourcePath: string,
       ): Promise<SourceMapInput | null> => {
-        if (parsedSourceMapCache.has(sourcePath)) {
-          return parsedSourceMapCache.get(sourcePath)!;
-        }
-
         if (browserResolveSourcemap) {
           const resolved = await browserResolveSourcemap(sourcePath);
           if (resolved.handled) {
-            parsedSourceMapCache.set(sourcePath, resolved.sourcemap);
             return resolved.sourcemap;
           }
         }
 
         const getSourceMaps = nodeResourceByAssetName.get(sourcePath);
         const sourceMap = (await getSourceMaps?.([sourcePath]))?.[sourcePath];
-        const parsed = sourceMap ? JSON.parse(sourceMap) : null;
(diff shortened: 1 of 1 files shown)


packages/core/src/core/runTests.ts
```diff
@@ -430,19 +426,30 @@ export async function runTests(context: Rstest): Promise<void> {
         }
       }
 
+      // Memoize parsed sourcemaps to avoid redundant JSON.parse when the same
+      // file is referenced by multiple errors or stack frames.
+      const parsedSourceMapCache = new Map<string, SourceMapInput | null>();
```
9aoy: Caching too much content can lead to excessive memory pressure, which is especially important to watch out for in large projects. Additionally, some content may only be accessed once.

We need a more flexible caching strategy—perhaps something like LRU?

fi3ework: make sense to me, it's not hot path so i will simply revert the change here to avoid the complex implementation.

PR #1083, opened 2026-03-23
feat: add `output.distPath.root` configuration

9aoy:
## Summary

Add the `output.distPath.root` configuration to control the global root directory where rstest places its temporary build outputs.

- **Type:** `string | { root?: string }`
- **Default:** `{ root: 'dist/.rstest-temp' }`

By default, rstest writes its temporary outputs under `dist/.rstest-temp` when they need to be written to disk. This includes the compiled test outputs used by the Node.js test runner, and the browser mode temporary assets such as the generated runner files and virtual manifest.

## Checklist

### DIFF
--- a/website/docs/zh/config/build/output.mdx
+++ b/website/docs/zh/config/build/output.mdx
@@ -147,16 +147,22 @@ export default defineConfig({
 
 用于自定义 CSS Modules 的配置。
 
+## output.cleanDistPath <RsbuildDocBadge path="/config/output/clean-dist-path" text="output.cleanDistPath" />
+
+是否在测试开始前，清空输出目录下的所有测试临时文件。
+
+默认情况下，Rstest 不会将测试临时文件写入磁盘，当你开启 Rstest 产物调试时可能需要此配置项。
+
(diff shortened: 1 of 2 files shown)


website/docs/zh/config/build/output.mdx
```diff
@@ -147,6 +147,36 @@ export default defineConfig({
...
 用于自定义 CSS Modules 的配置。
 
+## output.distPath <RsbuildDocBadge path="/config/output/dist-path" text="output.distPath" />
+
+- **类型：** `string | { root?: string }`
+- **默认值：** `{ root: 'dist/.rstest-temp' }`
+
+控制 rstest 临时构建产物的全局输出根目录。
+
+默认情况下，当 rstest 需要把临时产物写入磁盘时，会输出到 `dist/.rstest-temp`。这包括 Node.js 测试运行时使用的编译产物，也包括 browser mode 下生成的 runner 文件、virtual manifest 等临时资源。
```
fi3ework: do we, i think we only write to disk in debug mode.

PR #1088, opened 2026-03-23
fix(core): pin pool worker node runtime to parent execPath

omarshibli:
## Problem
`rstest` worker pools did not explicitly pin `execPath`, so worker runtime selection could drift based on environment/PATH resolution.
That can cause parent/worker Node-version divergence and runtime-specific failures.

## Root Cause
Both pool creation points relied on implicit process spawning defaults instead of forcing workers to use the parent process’ Node binary.

## Fix
Set `execPath: process.execPath` in both pool constructors:
- Forks pool
- Global setup pool

## Files Changed
- `packages/core/src/pool/forks.ts`
- `packages/core/src/core/globalSetup.ts`
- `packages/core/tests/pool/forksExecPath.test.ts`
- `packages/core/tests/core/globalSetupExecPath.test.ts`

## Verification
1. Added regression tests first to assert both pools pass `execPath: process.execPath`.
2. Confirmed red -> green:
- With fix removed: new tests fail.
- With fix restored: new tests pass.
3. Latest run:
- `pnpm test packages/core/tests/pool/forksExecPath.test.ts packages/core/tests/core/globalSetupExecPath.test.ts`
- Result: `2 passed, 0 failed`.

### DIFF
--- a/packages/core/src/core/globalSetup.ts
+++ b/packages/core/src/core/globalSetup.ts
@@ -22,6 +22,7 @@ const __dirname = dirname(__filename);
 function createSetupPool() {
   const options: Options = {
     runtime: 'child_process',
+    execPath: process.execPath,
     filename: resolve(__dirname, './globalSetupWorker.js'),
     execArgv: [
       ...process.execArgv,
--- a/packages/core/src/pool/forks.ts
+++ b/packages/core/src/pool/forks.ts
@@ -93,6 +93,7 @@ export const createForksPool = (poolOptions: {
 
   const options: Options = {
     runtime: 'child_process',
+    execPath: process.execPath,
     filename: resolve(__dirname, './worker.js'),
     env,
     execArgv,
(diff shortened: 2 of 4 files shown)


fi3ework:
1. **Pre-#1171 (tinypool):** tinypool's `Options` type doesn't declare `execPath`, and tinypool's internal `worker.initialize({...})` forwards a hardcoded option whitelist that excludes `execPath`. The property is dropped before reaching `child_process.fork`. With or without this PR, worker `process.execPath` always equals parent's (verified by spawning Node 20 with PATH-first Node 24 — worker stayed on Node 20 without the fix).
2. **Post-#1171 (current main):** we now call `node:child_process.fork` directly. `fork()`'s `execPath` defaults to `process.execPath` per Node spec, so the same invariant holds.
3. The provided regression tests assert on the tinypool constructor's argument shape, not on the worker's actual runtime — they wouldn't have caught either the tinypool whitelist or any real drift.

Closing as obsolete. Happy to take a follow-up that adds an explicit `execPath: process.execPath` (defensive/intent-documenting) plus a real end-to-end test that asserts `worker.process.execPath === parent.process.execPath` against the new pool, if you'd like to take another pass.

omarshibli:
Thanks for the follow up.

PR #1118, opened 2026-03-30
feat(core): support bundleDependencies patterns

9aoy:
### Background
`output.bundleDependencies` only supported a boolean toggle, so users could not inline a specific dependency or target package subpaths the way Vitest's inline dependency config does.

### Implementation
- Extend `output.bundleDependencies` to accept string and `RegExp` patterns for package names, subpaths, and glob-like `*` matches.
- Update the externalization plugin to match both direct requests and resolved `node_modules` specifiers, while keeping relative imports inside matched dependencies bundled.
- Add e2e coverage for package-name, subpath, glob, and relative-import cases, and document the new config forms in English and Chinese docs.

### User Impact
Users can now selectively bundle dependencies with entries like `['strip-ansi']`, `['strip-ansi/lib/index.js']`, `['strip-ansi/*']`, or `[/^strip-ansi\\/lib\\//]`.

## Checklist

### DIFF
--- a/website/docs/zh/config/build/output.mdx
+++ b/website/docs/zh/config/build/output.mdx
@@ -225,6 +225,11 @@ export default defineConfig({
 - `bundleDependencies: true` + `externals: ['lodash']`：打包所有依赖，但外部化 `lodash`。
 - `bundleDependencies: false` + `externals` 无需配置，因为所有依赖已被外部化。
 
+实际使用时，这两个配置更适合相反的场景：
+
+- 当大部分依赖都应该保持 external，只有少量包或子路径需要被打包时，更适合使用 `bundleDependencies` pattern。
+- 当大部分依赖都应该保持打包，只有少量包需要被 externalize 时，更适合使用 `output.externals`。
+
 ## output.cssModules <RsbuildDocBadge path="/config/output/css-modules" text="output.cssModules" />
 
 用于自定义 CSS Modules 的配置。
--- a/website/docs/zh/config/build/output.mdx
+++ b/website/docs/zh/config/build/output.mdx
@@ -148,13 +148,17 @@ export default defineConfig({
 
 <ApiMeta addedVersion="0.9.5" />
 
-- **类型：** `boolean`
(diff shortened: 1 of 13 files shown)


website/docs/zh/config/build/output.mdx
```diff
@@ -148,13 +148,17 @@ export default defineConfig({
...
 - **默认值：** 取决于 `testEnvironment`
 
 控制是否打包 `node_modules` 中的第三方依赖。
 
 - `true`：无论测试环境如何，始终打包所有第三方依赖。
 - `false`：无论测试环境如何，始终外部化第三方依赖。
+- `['pkg-name']`：仅打包列出的包，其余依赖保持外部化。
+- `['pkg-name/subpath']`：仅打包指定的包子路径。
+- `['pkg-name/*']`：按类似 glob 的模式匹配并打包包请求。
+- `[/^pkg-name\\/subpath/]`：按正则表达式匹配并打包包请求。
```
fi3ework, line 161: what is the difference between supporting string and RegExp and `output.external`?

website/docs/zh/config/build/output.mdx
```diff
@@ -188,13 +192,44 @@ export default defineConfig({
...
+import { defineConfig } from '@rstest/core';
+
+export default defineConfig({
+  output: {
+    bundleDependencies: ['strip-ansi/lib/index.js', 'strip-ansi/*'],
+  },
+});
+```
+
+这些模式只会影响 rstest 在当前构建图里仍然能够处理到的依赖请求。如果某个包本身已经被 externalize 了，那么它内部继续引用的子依赖也会保持 external。也就是说，`bundleDependencies` 不会把仅通过某个已 externalize 父包间接到达的传递依赖重新打包回来。
```
fi3ework, line 219: emm, i can't understand this section tbh.

9aoy: Updated. `bundleDependencies` only works for dependencies that rstest still handles during bundling. If a parent package is already externalized, rstest will not continue bundling that package's internal imports, so its child dependencies will stay external too.

PR #1152, opened 2026-04-09
fix: use web condition names by default in browser-like test environment

9aoy:
### Background
browser-like test environments were still resolving with node-oriented defaults, which makes browser-field reproduction harder and left this regression without a framework example outside the existing React and Vue samples.

### Implementation
- Update the core Rsbuild plugin to default web-like environments to `['browser', '...']` condition names. (follow https://github.com/web-infra-dev/rspack/blob/7165b2395e37c73732c5ede3664ca382faae13d0/packages/rspack/src/config/defaults.ts#L1104)
- Add an explicit core test for the default jsdom `conditionNames` behavior and refresh the affected snapshots.
- Add a new `examples/svelte-rsbuild` workspace with adapter-based config, minimal Svelte component tests, and lockfile updates.

### User Impact
Web-like projects now resolve browser-targeted exports by default, and the repo includes a Svelte Rsbuild example that reproduces and validates this path.

## Related Links

https://github.com/web-infra-dev/rstest/pull/1151

## Checklist

### DIFF
--- a/packages/core/src/core/plugins/basic.ts
+++ b/packages/core/src/core/plugins/basic.ts
@@ -68,9 +68,7 @@ export const pluginBasic: (context: RstestContext) => RsbuildPlugin = (
           resolve: {
             // override default resolve conditionNames for browser-like environment, use `browser` field instead of `node` field
             conditionNames:
-              testEnvironment.name === 'node'
-                ? undefined
-                : ['webpack', 'development', 'browser', '...'],
+              testEnvironment.name === 'node' ? undefined : ['browser', '...'],
           },
           output: {
             assetPrefix: '',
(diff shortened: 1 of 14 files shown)


packages/core/src/core/plugins/basic.ts
```diff
@@ -65,6 +65,13 @@ export const pluginBasic: (context: RstestContext) => RsbuildPlugin = (
               'import.meta.env': 'process.env',
             },
           },
+          resolve: {
+            // override default resolve conditionNames for browser-like environment, use `browser` field instead of `node` field
+            conditionNames:
+              testEnvironment.name === 'node'
+                ? undefined
+                : ['webpack', 'development', 'browser', '...'],
```
fi3ework: is `development` necessary since we're already in `development` mode and it will be inherited from `...`

9aoy: 👍

fi3ework (approved):

PR #1192, opened 2026-04-19
docs(migration): add shim/test-name guidance and snapshot format notes

fi3ework:
### Background

The `migrate-to-rstest` agent skill carried migration guidance (no-shim rule, test-name preservation, Jest snapshot key separator change, Vitest `mergeConfig` absence) that wasn't in the official docs. Moving it here makes the Jest/Vitest migration pages the single source of truth.

### Implementation

- Add **Don't shim — rewrite call sites** and **Preserve test names during replace** subsections to both Jest and Vitest guides
- Add **Snapshot format** section to the Jest guide covering the `:` → ` > ` key separator change, re-record workflow, and diff-review tips
- Add a `mergeConfig is not exported` tip before the Vitest config mapping table
- Mirror all additions in `docs/zh/`

### User Impact

Users migrating from Jest/Vitest get the full guidance inline without needing the agent skill.

---

Validation: docs-only change — ran `pnpm prettier --write`, `pnpm heading-case`, and `pnpm --filter rstest-website build` (all pass). Skipped full test suites.


website/docs/en/guide/migration/jest.mdx
```diff
@@ -177,6 +177,29 @@ Rstest provides a `rstest` API that you can use to access Rstest's utilities, su
...
+// Avoid: local rebinding
+const jest = rstest;
+
+// Avoid: aliased named import
+import { rstest as jest } from '@rstest/core';
+```
+
+The test body still reads `jest.fn()` after these, so a reviewer cannot tell the migration happened, IDE rename refactors stop following the new identifier, and any future `@rstest/core` deprecation warning will target an identifier the test never actually uses. The correct form is the literal rewrite shown above: every `jest.<api>` call site becomes `rstest.<api>`.
+
+#### Preserve test names during replace
```
9aoy: I’m a bit concerned about the positioning of this migration doc.

A large part of it reads more like a machine-consumable migration spec rather than a human-friendly guide.

Should we consider moving some of these parts into the migrate-to-rstest skill, and keep the doc more concise for humans?
Alternatively, some sections might be better kept as MD-only (agent-facing) content instead of being fully rendered. https://rspress.rs/guide/basic/ssg-md#how-to-implement-ssg-md:~:text=Provides%20import.meta.env.SSG_MD%20environment%20variable%2C

fi3ework: i am working with the migrate-to-rstest skill to rewrite the migration documentation. the migration documentation will be consumed by both humans and agents, and agents may even skip the skill and consume the migration documentation directly (users may not use the skill for some reason), so based on the above facts, i am modifying the migration documentation and skill to the following paradigm:

1. the migration documentation will only cover migration mapping and will not include how-to guides.
2. the skill only covers the text in the action guide and tries not to touch migration mapping. we treat the migration documentation as the single source of truth for migration mapping. the skill only assists the agent in writing and constrains red-line behavior, letting the agent know the direction of action for the harness project.

the current comment's code is outdated, you can review the latest code again.

fi3ework: i have significantly streamlined the content of the skill. you can see it here: https://github.com/rstackjs/agent-skills/commit/7fd6248fc4f9279019fd7197a56e2f016636bfdb

website/docs/en/guide/migration/vitest.mdx
```diff
@@ -80,13 +80,28 @@ When migrating config, keep two changes in mind:
 
 Walk through **every** option under `test` and match it against the table below — move it to the top level, rename it, or drop it. Fields not listed here may not map 1:1; verify against the [Rstest config reference](/config) before dropping them silently.
 
+:::tip `mergeConfig` is not exported
+Vitest's `mergeConfig` helper is not re-exported from `@rstest/core`. If you compose configs (for example, a shared base + per-package overrides), use plain object spread inside `defineConfig`:
```
9aoy: Use `mergeRstestConfig`  https://rstest.rs/api/javascript-api/rstest-core#mergerstestconfig

9aoy (approved):

PR #1196, opened 2026-04-21
feat(core): support build cache

9aoy:
## Summary

This PR adds `performance.buildCache` support for rstest test builds with rstest-aware defaults, adapter mapping, browser-mode integration, and dedicated coverage for warm-cache behavior. It also documents the override vs extend semantics, adds build/browser e2e coverage, and includes a smaller `virtual-module` build-cache regression case to help narrow a warm-run mock failure seen in `externals/mocked.test.ts`.

Benchmark results (antd + three.js example):

Run | Build | Total
-- | -- | --
Cold | ~385 ms | ~2.1 s
Warm | ~182 ms | ~1.7 s

≈2× faster build on warm runs

## Related Links

- https://github.com/web-infra-dev/rspack/pull/13775

## Checklist


website/docs/zh/config/build/performance.mdx
```diff
@@ -0,0 +1,66 @@
+---
+description: 配置 rstest 在测试构建中使用的 Rsbuild performance 选项。
+---
+
+import { RsbuildDocBadge } from '@components/RsbuildDocBadge';
+import { ApiMeta } from '@components/ApiMeta';
+
+# performance
+
+## performance.buildCache <RsbuildDocBadge path="/config/performance/build-cache" text="performance.buildCache" />
```
fi3ework, line 10: we need to add the similar https://rsbuild.rs/zh/config/performance/build-cache `:::tip` of why cache is not enabled by default. since we didn't documented any caveat of it.

packages/adapter-rsbuild/src/index.ts
```diff
@@ -25,6 +26,41 @@ export interface WithRsbuildConfigOptions {
   modifyRsbuildConfig?: (buildConfig: RsbuildConfig) => RsbuildConfig;
 }
 
+const addRsbuildConfigDependency = (
```
fi3ework: does Rslib also need to do the same staff, to count rslib config file into the digest calc?

fi3ework (commented):
what about `adapter-rspack`?

fi3ework (approved):

Issue #358, opened 2025-07-04
[Bug]: `importDynamic` parser flag should respect `importFunctionName`


fi3ework:
### Version

```sh
0.0.5
```

### Details

should not skipping the parser hook by using `importDynamic`, but rather using an early hook to simply respect `importFunctionName`.

### Reproduce link

Rslib test

### Reproduce Steps

Check out Rslib output `tests/dist/.rstest-temp/integration_bundle-false_index~test~ts.js`

PR #1210, opened 2026-04-28
feat(core): resolve dynamic imports against source module origin

### LINKED ISSUE #1207: [Bug]:  Dynamic `import()` with template literal resolves relative path against test file instead of source module
### Version

```sh
System:
    OS: macOS 26.4.1
    CPU: (8) arm64 Apple M1 Pro
    Memory: 225.70 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Browsers:
    Edge: 147.0.3912.86
    Safari: 26.4
  npmPackages:
    @rstest/core: ^0.9.9 => 0.9.9
```

## Description

When a bundled dependency contains a dynamic `import()` with a template literal (e.g., `` import(`./translations/${locale}/strings.json`) ``), rstest resolves the relative path against the **test file's location** instead of the **source module's location**. This causes `ERR_MODULE_NOT_FOUND` for files that exist in the dependency but not next to the test file.

This is a common pattern for i18n/localization — loading locale-specific JSON at runtime — and is a migration blocker from Jest for monorepos that use it.

## Reproduction

A dependency exports a function that does a dynamic import with a template literal:

```js
// lib-a/index.js
export const fetchStrings = (locale) =>
  import(`./translations/${locale}/strings.json`);
```

```json
// lib-a/translations/en-us/strings.json
{ "greeting": "Hello" }
```

A test in a different package imports and calls it:

```js
// app-b/src/app.test.js
import { fetchStrings } from 'lib-a';

test('fetchStrings resolves the JSON from lib-a', async () => {
  const strings = await fetchStrings('en-us');
  expect(strings.greeting).toBe('Hello');
});
```

### Failing e2e test

I've prepared a branch with a failing e2e test that follows the existing `e2e/externals/` conventions:

**Branch:** [`fix/dynamic-import-relative-path-e2e`](https://github.com/christiango/rstest/tree/fix/dynamic-import-relative-path-e2e)
**Diff:** [`main...fix/dynamic-import-relative-path-e2e`](https://github.com/christiango/rstest/compare/main...fix/dynamic-import-relative-path-e2e)

To run it locally:
```bash
git remote add christiango https://github.com/christiango/rstest.git
git fetch christiango fix/dynamic-import-relative-path-e2e
git checkout christiango/fix/dynamic-import-relative-path-e2e
pnpm install
cd e2e && pnpm rstest run externals/dynamicImportRelative.test.ts
```

## Expected behavior

`./translations/en-us/strings.json` resolves relative to `lib-a/index.js` (the module containing the `import()` call), finding `lib-a/translations/en-us/strings.json`.

## Actual behavior

rstest resolves the path relative to the **test file** (`app-b/src/app.test.js`), producing:

```
Error: Cannot find module '.../app-b/src/translations/en-us/strings.json'
    imported from .../node_modules/@rstest/core/dist/0~loadEsModule.js
```

## Root cause analysis

This appears to be a **known issue** — there's a TODO comment in the source at [`packages/core/src/runtime/worker/loadEsModule.ts` line 105](https://github.com/web-infra-dev/rstest/blob/main/packages/core/src/runtime/worker/loadEsModule.ts#L105):

```ts
// TODO: use module path instead of testPath
import.meta.resolve(specifier, pathToFileURL(testPath));
```

When `defineRstestDynamicImport` encounters a relative specifier that isn't in `assetFiles` (which is the case for template-literal dynamic imports since rspack can't statically resolve them), it falls back to `import.meta.resolve(specifier, pathToFileURL(testPath))`. This resolves the path relative to the **test file** instead of the **originating dependency module**.

Because rspack bundles everything into a single file, the originating module's identity is lost by the time the dynamic import runs at runtime. The `importModuleDynamically` callback's `_referencer` parameter points to the single bundle module, not the original source file that contained the `import()` call. So a straightforward fix using the referencer won't work — the resolution context would need to be preserved through the bundling step, likely via rspack's context module handling for template-literal imports.

**Key source locations:**
- `defineRstestDynamicImport` fallback: [`loadEsModule.ts` L100-106](https://github.com/web-infra-dev/rstest/blob/main/packages/core/src/runtime/worker/loadEsModule.ts#L100-L106)
- `importModuleDynamically` callback: [`loadEsModule.ts` L278-289](https://github.com/web-infra-dev/rstest/blob/main/packages/core/src/runtime/worker/loadEsModule.ts#L278-L289)

### Reproduce link

https://github.com/christiango/rstest/compare/main...fix/dynamic-import-relative-path-e2e)

### Reproduce Steps

git remote add christiango https://github.com/christiango/rstest.git
git fetch christiango fix/dynamic-import-relative-path-e2e
git checkout christiango/fix/dynamic-import-relative-path-e2e
pnpm install
cd e2e && pnpm rstest run externals/dynamicImportRelative.test.ts

fi3ework:
## Summary

`RstestPlugin` now sets `injectDynamicImportOrigin: true`, and `defineRstestDynamicImport` (ESM + CJS) handles the third `origin` arg that rspack's matching rewrite (web-infra-dev/rspack#13849) appends to rewritten `import()` calls.

- `import.meta.resolve` uses `origin ?? testPath` as the base, so a template-literal `import()` inside a bundled dep resolves against the dep's own directory instead of the test entry's.
- When `origin` is set, the runtime returns Node's raw Module Namespace Object directly. The existing CJS interop `Proxy` is bypassed on this path: it doesn't define `ownKeys`, so `Object.keys(ns)` would miss `default` — observably different from the `SyntheticModule` namespace that string-literal `import()` produces today. String-literal specifiers and the vm `importModuleDynamically`/`link` callbacks (no `origin`) keep their existing interop path.

Fixes #1207.

## Companion change

Depends on **web-infra-dev/rspack#13849**. Until `@rspack/core` ships the option, the call-site rewrite is a no-op and the new e2e is `it.todo`. A single `@ts-expect-error` next to `injectDynamicImportOrigin: true` bridges the TS-level mismatch — when rspack ships the new binding type, that comment will error out and force removal.

## Test plan

- [x] `pnpm --filter @rstest/core test`
- [x] `pnpm --filter @rstest/core typecheck`, `pnpm typecheck`, `pnpm lint`
- [x] e2e externals — existing tests pass; new e2e is `it.todo` so CI isn't blocked
- [ ] After web-infra-dev/rspack#13849 ships: bump `@rspack/core`, switch e2e to `it`, drop the `@ts-expect-error`

Refs #1207, #358

### DIFF
--- a/e2e/dom/fixtures/package.json
+++ b/e2e/dom/fixtures/package.json
@@ -12,7 +12,7 @@
     "react-dom": "^19.2.5"
   },
   "devDependencies": {
-    "@rsbuild/core": "^2.0.3",
+    "@rsbuild/core": "^2.0.5",
     "@rsbuild/plugin-react": "^2.0.0",
     "@testing-library/dom": "^10.4.1",
     "@testing-library/jest-dom": "^6.9.1",
--- a/e2e/externals/dynamic-import-interop/babelStyle.cjs
+++ b/e2e/externals/dynamic-import-interop/babelStyle.cjs
@@ -1,0 +1,7 @@
+// Babel/TSC-style CJS: `__esModule: true` plus an explicit `exports.default`.
+// User contract: `import x from 'lib'` should give the inner default value
+// (not the whole `module.exports`).
+'use strict';
+Object.defineProperty(exports, '__esModule', { value: true });
+exports.a = 'world';
+exports.default = () => 'hello';
--- a/e2e/externals/dynamic-import-interop/plain.cjs
+++ b/e2e/externals/dynamic-import-interop/plain.cjs
@@ -1,0 +1,6 @@
+// Plain CJS without `__esModule`. User contract: `default` should be the
+// whole `module.exports`; named exports may be reachable via lexer detection
+// or default-fallback.
+'use strict';
+exports.foo = 'foo';
+exports.bar = 'bar';
--- a/e2e/externals/dynamic-import-interop/webpackStyle.cjs
+++ b/e2e/externals/dynamic-import-interop/webpackStyle.cjs
@@ -1,0 +1,7 @@
+// Webpack/rspack-style CJS bundle: `__esModule: true` is stamped via
+// `Object.defineProperty`, named exports are mirrored onto `exports`. This
+// is the rslib eco-ci format-snapshot shape.
+'use strict';
+Object.defineProperty(exports, '__esModule', { value: true });
+exports.bar = 'bar';
+exports.foo = 'foo';
--- a/e2e/externals/dynamic-import-origin/index.js
+++ b/e2e/externals/dynamic-import-origin/index.js
@@ -1,0 +1,5 @@
+// Template-literal dynamic import. Without `injectDynamicImportOrigin`,
+// rstest used to resolve the relative specifier against the test entry,
+// which sits in a different directory and does not contain `./translations`.
+export const fetchStrings = (locale) =>
+  import(`./translations/${locale}/strings.json`);
--- a/e2e/externals/dynamic-import-origin/translations/en-us/strings.json
+++ b/e2e/externals/dynamic-import-origin/translations/en-us/strings.json
@@ -1,0 +1,1 @@
+{ "greeting": "Hello" }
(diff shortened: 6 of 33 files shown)


fi3ework:
@codex review

fi3ework:
@codex review

PR #1171, opened 2026-04-14
feat(pool): replace tinypool with self-owned worker pool

fi3ework:
## Summary

Replace tinypool with a self-owned three-layer worker pool for full control over worker spawn, communication, error handling, and lifecycle management.

**Motivation:**

As rstest's user base grows and usage scenarios become more complex, we need deeper customization of the worker lifecycle — requirements that go beyond what tinypool was designed to provide as a general-purpose pool library. Specifically:

- Precise control over IPC send behavior during worker shutdown, with cross-platform error handling (#1142)
- A structured crash reporting channel instead of relying solely on stderr output (#1033)
- Built-in per-task stderr attribution so crash output enters the reporter rather than printing directly to the terminal (#956)
- First-class APIs for spawn events and worker identity, rather than depending on internal fields
- Spawn-on-demand lazy creation strategy, paving the way for memory-aware worker creation (#1160)

To meet these needs, we had accumulated several adaptation layers: `stderrCapture.ts` (328 lines), `workerMeta.ts` (51 lines), tinypool internal field access, and message filtering. As customization requirements continue to grow, a self-owned pool is the more sustainable path forward.

**Architecture:**

Three-layer design:

- **`Pool`** — Scheduling layer: max/min worker management, slot queuing, close drain
- **`PoolRunner`** — State machine layer: `IDLE → STARTING → STARTED → STOPPING → STOPPED`, birpc wiring, crash attribution (`crashed` / `isUsable()`), operation lock serialization
- **`ForksPoolWorker`** — Transport layer: `child_process.fork`, built-in stderr capture, IPC error tolerance

**Changes overview:**

- Added `pool.ts`, `poolRunner.ts`, `poolWorker.ts`, `protocol.ts`, `forksPoolWorker.ts`
- Removed `forks.ts`, `stderrCapture.ts`, `workerMeta.ts` and their adaptation logic
- Refactored worker entry: extracted `runInPool.ts`, simplified `worker/index.ts` to lifecycle bootstrap
- Removed `tinypool` dependency

**Acknowledgements:**

Thanks to tinypool for providing a stable and reliable worker pool foundation during rstest's early development, allowing us to focus on building the testing framework itself. This replacement is not a reflection of any shortcoming in tinypool — rather, rstest's worker lifecycle customization needs have grown beyond the scope of a general-purpose pool library.

## Test plan

[...]

PR #1216, opened 2026-04-30
fix(pool): restore stderr truncation and settle wait lost in #1171

fi3ework:
### Background

PR #1171 replaced tinypool with a self-owned worker pool and deleted `stderrCapture.ts`, internalizing its behavior into `ForksPoolWorker` and `PoolRunner`. Two behaviors from #956 were lost in the process: message-level stderr truncation (64 KB cap with head/tail marker) and a settle wait that lets pending stderr `data` events drain before reading the buffer on worker exit.

### Implementation

- Add `formatCapturedStderr` to `poolRunner.ts` — truncates stderr to 64 KB in error messages using head/tail halves with a `[truncated N bytes of stderr]` marker, matching the old `stderrCapture.ts` behavior.
- Add `waitForStderrSettle()` to the `PoolWorker` interface and `ForksPoolWorker` — waits for stderr stream `close` or a 200 ms timeout before the rejection path reads the buffer, closing the race window between the `exit` event and pending `data` callbacks.
- `rejectCurrentTaskWithStderr` now defers rejection through `waitForStderrSettle()` so crash output written immediately before `process.exit()` is captured.

### User Impact

Worker crash error messages are now bounded at ~64 KB instead of up to 1 MB, and stderr written right before a crash is less likely to be silently dropped.

## Related Links

- #1171 — pool rewrite that introduced the regression
- #956 — original fix that added stderr capture for crash diagnostics

## Checklist

### DIFF
--- a/packages/core/src/pool/poolRunner.ts
+++ b/packages/core/src/pool/poolRunner.ts
@@ -13,6 +13,19 @@ import type { PoolTask } from './types';
 
 const WORKER_START_TIMEOUT_MS = 90_000;
 const WORKER_STOP_TIMEOUT_MS = 60_000;
+const MAX_STDERR_MESSAGE_BYTES = 64 * 1024;
+
+function formatCapturedStderr(text: string): string {
+  const bytes = Buffer.byteLength(text);
+  if (bytes <= MAX_STDERR_MESSAGE_BYTES) {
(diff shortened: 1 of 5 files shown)


fi3ework:
@codex review

PR #1224, opened 2026-05-06
feat: support silent logs

9aoy:
## Summary

This PR adds support for `silent` test logs, including `silent: 'passed-only'`, so intercepted console output can be suppressed or replayed only for failed tasks. It also keeps `silent` working when `disableConsoleIntercept` is enabled, while preserving the expected behavior that `onConsoleLog` and `printConsoleTrace` stay disabled on that path.

 `silent` 
- **Type:** `boolean | 'passed-only'`
- **Default:** `false`
- **CLI:** `--silent [value]`

Silence intercepted console output from tests.

- Set `true` to hide all intercepted console logs.
- Set `'passed-only'` to keep intercepted logs only for failed files, suites, and test cases.

## Example

When `silent` is set to `'passed-only'`, console logs from passing tasks are buffered and discarded, while logs from failed tasks are replayed with the failure output.

```ts title="example.test.ts"
import { describe, expect, it } from '@rstest/core';

describe('math', () => {
  it('passes', () => {
    console.log('pass log');
    expect(1 + 1).toBe(2);
  });

  it('fails', () => {
    console.log('fail log');
    expect(1 + 1).toBe(3);
  });
});
```

Only `fail log` will be printed.

## Related Links

None.

## Checklist


9aoy:
@codex review

packages/browser/src/client/entry.ts
```diff
@@ -611,15 +633,39 @@ const run = async () => {
...
+        });
+        workerState.currentTask = taskStack[taskStack.length - 1];
         dispatchRunnerLifecycle('case-start', test);
       },
       onTestCaseResult: async (result) => {
+        if (taskStack[taskStack.length - 1]?.taskId === result.testId) {
+          taskStack.pop();
+          workerState.currentTask = taskStack[taskStack.length - 1];
+        }
         if (result.status === 'fail') {
```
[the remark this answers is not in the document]
9aoy: @copilot apply changes based on [this feedback](https://github.com/web-infra-dev/rstest/pull/1224#discussion_r3194553317)

website/docs/zh/config/test/silent.mdx
fi3ework, line 1: `silent.mdx` doesn't mention `disableConsoleIntercept` at all, while `disable-console-intercept.mdx` already references `silent`. Given the PR explicitly supports the combined case (there's dedicated e2e coverage for `--silent=passed-only --disableConsoleIntercept`), it'd be good to add a short section to `silent.mdx` describing how the two configs interact, so the cross-reference is symmetric.

website/docs/zh/guide/migration/vitest.mdx
```diff
@@ -124,6 +124,7 @@ Vitest 配置文件中使用的 helper 可以对应到 `@rstest/core` 导出的
...
 | `includeTaskLocation`         | [`includeTaskLocation`](/config/test/include-task-location)                                    |                                                                                                                                                                                                                                                                                                                                                                                                                                                         |
+| `silent`                      | [`silent`](/config/test/silent)                                                                | 与 Vitest 一样支持 `boolean \\                                                                                                                                                                                                                                                                                                                                                                                                                          | 'passed-only'`，用于控制被拦截的测试 console 输出。 |
```
fi3ework: weird `\\`

packages/core/src/pool/index.ts
fi3ework: should use `getFileTaskId` now.

fi3ework (approved):

PR #1239, opened 2026-05-11
feat(browser-react): add React 17 entry via `@rstest/browser-react/react17`

fi3ework:
### Background

`@rstest/browser-react` statically imports `react-dom/client`, which does not exist under React 17. React 17 users hit a hard resolution error today and have no workable entry point.

### Implementation

- Extract a shared `createRenderApi` factory so `render` / `renderHook` / `cleanup` are reused across React majors.
- Add new entries `@rstest/browser-react/react17` (auto-cleanup) and `@rstest/browser-react/react17/pure`, mounting via `ReactDOM.render` / `unmountComponentAtNode` and pulling `act` from `react-dom/test-utils`.
- Detect `React.unstable_act` so the default entry also keeps working under React 18.0–18.2.

### User Impact

React 17 projects can now use `@rstest/browser-react/react17` without bundler configuration. React 18+ users are unaffected (default entry is unchanged).

## Checklist

### DIFF
--- a/e2e/browser-mode/browserReact17.test.ts
+++ b/e2e/browser-mode/browserReact17.test.ts
@@ -1,0 +1,15 @@
+import { describe, expect, it } from '@rstest/core';
+import { runBrowserCli } from './utils';
+
+describe('browser mode - @rstest/browser-react (React 17)', () => {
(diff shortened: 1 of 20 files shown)


fi3ework:
close until we decide to do it in the future.

PR #1248, opened 2026-05-12
ci: split PR baseline and cron supplemental e2e coverage

fi3ework:
### Background

The e2e matrix currently fires too many jobs on every PR — all three scripts (`test`, `test:commonjs`, `test:no-isolate`) run across every PR node combination, costing 10 jobs even for trivial changes. The intended shape is a fast baseline on every PR, the full supplemental sweep on every merge to `main`, and the same full sweep as a final gate on release PRs.

Per @9aoy's feedback: a nightly cron's failure surfaces too late. Folding the supplemental sweep into the push-to-`main` trigger keeps the PR matrix tight while still catching any regression the baseline misses, with timely notification on merge.

### Implementation

Drive all three axes off `github.event_name` and a new `prepare.outputs.is_release_pr` (matched via `startsWith(github.event.pull_request.title, 'release:')`):

```plaintext
trigger      | node_version | test_script                  | exclude
-------------+--------------+------------------------------+-----------
PR (regular) | [20, 24]     | [test]                       | -
push to main | [20, 22, 24] | [test, commonjs, no-isolate] | test × 22
release PR   | [20, 22, 24] | [test, commonjs, no-isolate] | test × 22
```

- PR (regular): 4 jobs (baseline `test` only) — down from 10.
- push to main: 16 jobs (union — `test` × Node 22 excluded so the baseline footprint matches the PR matrix).
- release PR: same matrix as push, runs as a pre-merge gate.
- The `schedule:` trigger and its conditional branches in `lint` / `ut` / `e2e` are removed; push-to-`main` is now the single supplemental trigger.
- Note inlined near the matrix: Node 20 + `test:no-isolate` can hit upstream vm native crashes under worker reuse; tolerated on push / release-PR runs.

### User Impact

None — internal CI cost reduction and notification-latency improvement. PR e2e job count: 10 → 4. Supplemental regression coverage moves from "next nightly cron" to "immediately on merge to main".

## Validation

Scoped to the workflow YAML file — no runtime/test source touched. Full build/test/e2e suites skipped as not applicable. `prettier --check` passes on the file.

## Checklist

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -160,8 +160,10 @@ jobs:
       matrix:
         # run ut in macOS, as SWC cases will fail in Ubuntu CI
         os: [macos-14, windows-latest]
-        node_version: ${{ fromJson(github.event_name == 'schedule' && '["20","22","24"]' || '["20","24"]') }}
-        test_script: ${{ fromJson(github.event_name == 'schedule' && '["test:commonjs","test:no-isolate"]' || '["test"]') }}
+        node_version: ${{ fromJson((github.event_name == 'schedule' || startsWith(github.event.pull_request.title, 'release:')) && '["20","22","24"]' || '["20","24"]') }}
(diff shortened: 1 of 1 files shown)


9aoy:
What if the full test is triggered when the **main** pushed? I'm afraid we won't receive the notification in time during the scheduled task.

9aoy (approved):

PR #1252, opened 2026-05-13
feat(core): auto install missing runtime dependencies

9aoy:
### Background

Missing optional runtime packages can fail late when coverage providers or DOM test environments are loaded from worker or merge paths.

### Implementation

- Add a main-process dependency preflight for coverage providers and DOM environments.
- Prompt to install `@rstest/coverage-istanbul`, `jsdom`, and `happy-dom` without pinning them to the Rstest version.
- Move coverage install helpers into `coverage/install.ts` so `coverage/index.ts` only resolves and loads providers.

### User Impact

Users can accept an interactive prompt to install missing coverage or DOM environment packages before tests run.

image

## Related Links

None

## Checklist


9aoy:
@codex review

packages/core/src/core/envDependencies.ts
```diff
@@ -0,0 +1,76 @@
...
+    ) {
+      continue;
+    }
+
+    packages.set(packageName, environmentName);
+  }
+
+  for (const [packageName, environmentName] of packages) {
+    await installer(packageName, root, environmentName, options);
+  }
```
fi3ework, line 124: No `isPackageInstalled` recheck after `installer()` returns, unlike `ensureCoverageProviderInstalled`. In non-TTY (CI / agent), installer is a no-op, so tests proceed and fail late inside the worker when `jsdom`/`happy-dom` cannot be loaded. Adding a recheck-and-throw here would surface the failure early in the main process and match the coverage path.

fi3ework (commented):
Bare-name install resolves to `latest`. Since `@rstest/core` and `@rstest/coverage-istanbul` version independently with a `workspace:~` peer, latest may pull a provider that doesn't satisfy the installed core's peer range. Aligning the two packages to the same version with a `workspace:*` peer would make auto-installing latest always safe.

This is a change in release strategy; perhaps we can keep all packages in the current repository on the same version.

fi3ework (approved):

PR #1263, opened 2026-05-15
feat(core): support changed coverage

9aoy:
## Summary

This PR adds changed-file scoped coverage so users can collect coverage only for files changed in the current Git repository, either through `coverage.changed` config or `--coverage.changed [commit]`. It also makes `--changed` reuse its changed source file set for coverage when coverage is enabled and `coverage.changed` is not explicitly configured, while preserving full coverage for force reruns unless changed coverage is explicitly requested.

## Checklist

### DIFF
--- a/website/docs/zh/config/test/coverage.mdx
+++ b/website/docs/zh/config/test/coverage.mdx
@@ -187,10 +187,12 @@ export default defineConfig({
 npx rstest run --changed --coverage
 npx rstest run --coverage.changed
 npx rstest run --coverage.changed=HEAD~1
+npx rstest run --coverage.changed=origin/main
 ```
 
 - `npx rstest run --changed --coverage` 会运行变更相关测试，并且默认把覆盖率也限制到同一批变更文件。
 - `npx rstest run --coverage.changed=HEAD~1` 会执行正常测试集，但只报告相对 `HEAD~1` 以来变更文件的覆盖率。
+- `npx rstest run --coverage.changed=origin/main` 适合在功能分支上使用：它仍然会执行正常测试集，但覆盖率报告只包含相对 `origin/main` 变更过的文件。
 
 ### exclude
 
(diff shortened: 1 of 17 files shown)


9aoy:
@codex review

website/docs/zh/config/test/coverage.mdx
```diff
@@ -123,6 +124,34 @@ export default defineConfig({
...
+
+<ApiMeta addedVersion="0.10.0" />
+
+- **类型：** `boolean | string`
+- **默认值：** `undefined`
+- **CLI：** `--coverage.changed`, `--coverage.changed=<commit>`
+
+只收集当前 Git 仓库中变更文件的覆盖率。你也可以传入 commit 或 branch，收集从该 ref 以来变更文件的覆盖率。
+
+`coverage.changed` 只控制覆盖率报告范围，不会改变要运行的测试。当使用 `--changed` 且没有配置 `coverage.changed` 时，覆盖率报告会继承 `--changed` 收集到的变更文件。
```
fi3ework: link to the `--changed` page.

website/docs/zh/config/test/coverage.mdx
```diff
@@ -123,6 +124,34 @@ export default defineConfig({
...
+
+```ts title='rstest.config.ts'
+import { defineConfig } from '@rstest/core';
+
+export default defineConfig({
+  coverage: {
+    enabled: true,
+    changed: 'HEAD~1',
+  },
+});
```
fi3ework, line 183: what is the intended effect of this code example? i'm not very clear about this flag, i currently feel that this flag is more likely used only in the CLI?

9aoy: `coverage.changed` is meant to control coverage scope, not test selection.
The main use case is not only “make --changed also shrink coverage”, but more generally “decouple which tests we run from which files we report coverage for”.

Typical scenarios are:

- users expect `--changed + --coverage` to report changed-file coverage instead of full-repo coverage;
- pre-commit / lint-staged workflows want coverage only for staged or changed files;
- CI may still run the full test suite, but only report coverage for PR-touched files, without running a second changed-only test pass.

So the config form is intentional: it lets users persist this as a coverage policy in project config, especially for CI, while the CLI flag is just the ad-hoc entry point.

fi3ework (approved):

Issue #1274, opened 2026-05-18
[Bug]: 0.10.0 is 30% slower, when using @rsbuild/plugin-babel


pkasarda:
### Version

```
@rstest/core: 0.9.9 → 0.9.10 → 0.10.0
Node:         v24
OS:           macOS arm64 (local) / Linux arm64 (CI)
```

### Setup

- Nx monorepo with many lib projects
- ~350 test files, ~3350 tests, mix of jsdom (React + Testing Library + MSW v1) and node-env tests
- Two run modes both reproduce the regression:
  1. **Workspace mode** — one rstest process across all projects (`projects: ['./libs/**/rstest.config.mts']` glob)
  2. **Per-project** — invoking one rstest process per project (the Nx execution model)

### Numbers

**CI** (4-core ARM runner, fresh checkout, isolated jobs, default `pool.maxWorkers = numCpus - 1`):

| Version | Wall | Δ vs 0.9.9 |
|---|---|---|
| 0.9.9  | 7:03 | baseline |
| 0.9.10 | 7:37 (avg of 7:40, 7:35) | +8% |
| 0.10.0 | 9:09 (avg of 9:29, 8:49, 9:29) | **+30%** |

**Local per-project** (one process per lib, 12-core ARM, format `wall (rstest internal)`):

| Version | small lib (38 tests, jsdom) | larger lib (191 tests, jsdom) |
|---|---|---|
| 0.9.9  | 2.93s (2.19s) | 6.23s (5.42s) |
| 0.9.10 | 4.74s (3.97s) | 12.87s (11.24s) |
| 0.10.0 | 4.96s (3.14s) | 15.24s (13.63s) |
| Δ 0.10.0 vs 0.9.9 | **+69%** | **+145%** |

The per-project regression is dramatically larger than workspace mode (+145% vs +30%), which suggests the overhead is concentrated in rstest's **startup / initialization**, not in test execution itself. Workspace mode pays the cost once; the per-project execution model pays it N times.

A patch-level regression already starts in `0.9.10` (+8% workspace, +60% per-project on the small lib), then 0.10.0 adds more on top.

### What changed between 0.9.9 → 0.9.10 → 0.10.0?

Without an rstest-side bisect tool I can't pinpoint it from outside. Would something like an rsdoctor-equivalent for rstest help here? Or could a maintainer run `git bisect` against your own benchmark suite and see if the regression reproduces?

### Reproduce

[...]

PR #1276, opened 2026-05-18
fix(core): only merge performance.buildCache when user opts in

pkasarda:
## Summary

`pluginBasic` writes `performance.buildCache` into the merged Rsbuild environment config on every run. When the user has **not** configured `performance.buildCache`, `resolveProjectBuildCache` returns `false`, so `performance.buildCache: false` is written explicitly. Rsbuild propagates that to Rspack as `cache: false`, which disables Rspack's default in-memory module cache (`cache: { type: 'memory' }`). Every build then starts cold from disk — roughly a **4× build-phase slowdown** on non-trivial test setups.

This PR guards the merge so `performance.buildCache` is only set when the resolved value is truthy. When the user opts in (`performance.buildCache: true | { ... }`), behaviour is unchanged.

Refs #1274.

## Bisect

Cold runs against a real-world Rsbuild + SWC + jsdom test config (191 jsdom tests, React/Lingui/Babel-macro pipeline). Build-phase only, median of N cold runs per commit:

| Commit | Build phase | Δ |
|---|---|---|
| `ff3d9a35` (parent of #1196) | **0.94s** | baseline |
| `c9907eb0` feat(core): support build cache (#1196) | **4.95s** | **+426%** |
| `2690a8bb` v0.10.0 | 5.32s | + small drift |

PR #1196 is the single cause. The other commits between v0.9.10 and v0.10.0 do not contribute.

## Hyperfine results

`hyperfine --warmup 1 --runs 10 --prepare 'rm -rf node_modules/.cache .rstest tmp/rstest'`, total wall time per run:

| Version | Mean ± σ | Min | Max | Relative |
|---|---|---|---|---|
| 0.9.10 | 7.55s ± 2.04s | 5.72s | 12.61s | 1.04× |
| 0.10.0 | 13.73s ± 3.07s | 10.60s | 18.76s | **1.89× slower** |
| 0.10.0 + this patch | **7.27s ± 0.92s** | 6.27s | 8.74s | **1.00×** |

Patched 0.10.0 is **1.89× faster than unpatched 0.10.0**, ties 0.9.10 within noise, with tighter variance (σ 0.92s vs 3.07s).

## Root cause

[`packages/core/src/core/plugins/basic.ts`](https://github.com/web-infra-dev/rstest/blob/main/packages/core/src/core/plugins/basic.ts) before:

```ts
return mergeEnvironmentConfig(
  config,
  {
    performance: {
      buildCache: resolveProjectBuildCache({ context, project }),
    },
    tools, resolve, source, output, dev,
  },
  // ...
);
```

`resolveProjectBuildCache` returns `false | RstestBuildCacheConfig`. When the user has not configured `performance.buildCache`, it returns `false`, and the literal above sets `performance.buildCache: false`. Rsbuild forwards `performance.buildCache: false` to Rspack as `cache: false`, disabling Rspack's default `cache: { type: 'memory' }`.

After this patch, `performance.buildCache` is only set in the merged config when truthy, so Rspack's in-memory cache default is preserved on opt-out users.

## Test plan

- [x] `pnpm --filter @rstest/core test` — passes
- [x] Manual bench above shows v0.10.0 + patch ties v0.9.10
- [ ] Existing `e2e/build/buildCache.test.ts` should continue to pass (it explicitly sets `performance.buildCache`, so this PR doesn't change its code path)
- [ ] If maintainers want, happy to add a regression test asserting the merged env config doesn't include `performance.buildCache` when the user has not opted in

Marked **draft** — let me know if you'd like the regression test before promoting it.

cc @9aoy @fi3ework

### DIFF
--- a/packages/core/src/core/plugins/basic.ts
+++ b/packages/core/src/core/plugins/basic.ts
@@ -50,15 +50,21 @@ export const pluginBasic: (context: RstestContext) => RsbuildPlugin = (
         multipleProjects: context.projects.length > 1,
       });
 
+      // Mirrors the pattern in packages/browser/src/hostController.ts:
+      // only write `performance.buildCache` when the user has opted in.
+      // Writing `performance.buildCache: false` explicitly causes Rsbuild
+      // to set Rspack's `cache: false`, which disables Rspack's default
+      // in-memory module cache and makes every build start cold from disk
+      // (~4x slowdown on non-trivial test builds).
+      const buildCache = resolveProjectBuildCache({
+        context,
+        project,
+      });
+
       return mergeEnvironmentConfig(
         config,
         {
-          performance: {
-            buildCache: resolveProjectBuildCache({
-              context,
-              project,
-            }),
-          },
+          performance: buildCache ? { buildCache } : undefined,
           tools,
           resolve,
           source,


pkasarda:
**CI bench results from applying this patch via `patch-package` on a real test suite.**

4-job CI matrix on the same workflow (cold GitHub Actions runners, no cache restore between jobs, ~3400 tests across ~350 files, rstest workspace mode). Numbers below are from rstest's own `Duration` line, not the workflow wall time.

| Job | total | build | tests |
|---|---|---|---|
| 0.9.9 vanilla | 9m 9s | 1m 21s | 7m 47s |
| 0.9.10 vanilla *(baseline)* | **6m 42s** | 58.9s | 5m 43s |
| 0.10.0 vanilla *(this branch's `main`)* | 9m 58s | 1m 21s | **8m 37s** |
| **0.10.0 + this patch** | **7m 4s** | **56.7s** | 6m 7s |

This patch saves **~3 minutes per CI run** vs vanilla 0.10.0 on this suite, and lands within ~5% of 0.9.10. Worth highlighting: the bug regresses the **tests phase** too, not just `build` (likely because rspack's in-memory module cache is used during per-worker compilation of dynamic imports / lazy chunks). The patched variant has the fastest build phase of all four — confirming that the only impact of the unconditional `performance.buildCache: false` write was to disable rspack's `cache: { type: 'memory' }` default.

Local hyperfine (8 cold runs, smaller fixture — 191 tests in a single lib):

| | Mean ± σ | Relative |
|---|---|---|
| 0.9.10 vanilla | 7.91s ± 2.07 | 1.12× |
| 0.10.0 vanilla | 10.64s ± 0.46 | 1.51× |
| 0.10.0 + `buildCache: true` workaround | 11.84s ± 2.73 | 1.68× (worst) |
| **0.10.0 + this patch** | **7.07s ± 0.92** | **1.00× (best)** |

The `buildCache: true` workaround is *worse* than the bug on cold cache, by the way — persistent-cache write overhead with zero hits. So this patch is the only viable fix for CI users.

Marking ready for review.

9aoy (approved):

9aoy:
nice catch 👍

PR #1287, opened 2026-05-20
feat(core): support custom test environments

9aoy:
## Summary

This PR adds config-level custom test environment support to `@rstest/core`, so projects are no longer limited to the built-in `node`, `jsdom`, and `happy-dom` environments. It allows loading environment objects from local JavaScript files or `rstest-environment-*` packages, exports `builtinEnvironments` and `TestEnvironment` types for user extension, clarifies when to use custom environments versus `setupFiles`, and adds unit and e2e coverage for builtin wrappers, named exports, and package fallback resolution.

```js title="my-environment.mjs"
import { builtinEnvironments } from '@rstest/core';

/** @type {import('@rstest/core').TestEnvironment<typeof globalThis, { marker: string }>} */
const environment = {
  name: 'custom-jsdom',
  async setup(global, options) {
    const base = await builtinEnvironments.jsdom.setup(global, {
      url: 'https://example.com',
    });

    global.__MARKER__ = options.marker;

    return {
      async teardown() {
        delete global.__MARKER__;
        await base.teardown();
      },
    };
  },
};

export default environment;
```

```ts title="rstest.config.ts"

export default defineConfig({
  testEnvironment: {
    name: './my-environment.mjs',
    options: {
      marker: 'custom-marker',
    },
  },
});
```

## Checklist


fi3ework (commented):
are the API of custom envs the final version before 1.0.0?

9aoy:
> are the API of custom envs the final version before 1.0.0?

@fi3ework  Not final yet. The current design should already cover the main use cases we want to support, but we may make minor adjustments based on real-world feedback.

website/docs/zh/config/test/test-environment.mdx
```diff
@@ -90,6 +96,208 @@ export default defineConfig({
...
+    },
+  },
+});
+```
+
+`transformMode` 本身不会创建 DOM API，也不会启用 [browser 模式](/config/test/browser)。它只选择默认的转换和解析行为。environment 实现本身仍然需要负责安装 `window`、`document` 等全局对象。
+
+Rstest 暂不支持文件级 `@vitest-environment` 或 `@jest-environment` 注释。如果你需要按文件切换环境，当前建议用不同的 project 或不同的配置文件来拆分。
+
+#### 从 Jest 或 Vitest 迁移自定义 environment \{#migrating-custom-environments}
```
fi3ework: what if move this to the migration page? it's quite detailed. the migration page has always been the single source of truth manual for migration.

website/docs/zh/config/test/test-environment.mdx
```diff
@@ -90,6 +96,208 @@ export default defineConfig({
...
+    },
+  },
+});
+```
+
+`@rstest/core` 导出了 `builtinEnvironments`，这样你可以基于 `node`、`jsdom` 或 `happy-dom` 做扩展，而不是从零实现。
+
+Rstest 当前会按 JavaScript module 加载自定义 environment 文件，因此暂不支持直接传 `.ts` custom environment 文件路径。如果你需要类型检查，可以先用 TypeScript 编写，再编译成 `.mjs` 或 `.js`，或者在 JavaScript 文件里使用 JSDoc 标注类型。
+
+#### transformMode
```
fi3ework: show the default value and type like other API fields.

website/docs/zh/config/test/test-environment.mdx
```diff
@@ -90,6 +96,208 @@ export default defineConfig({
...
+
+`transformMode` 用于控制自定义 environment 是否应该使用类 Node 或类浏览器的默认转换和解析行为。
+
+自定义 environment 默认使用 `transformMode: 'node'`，除非你显式配置。内置 environment 会自动推断该值，所以通常只有配置自定义 environment 时才需要使用这个选项。
+
+当 environment 使用 Node.js module 语义运行时，使用 `transformMode: 'node'`。在这个模式下，Rstest 会保留面向 Node 的解析默认值，并默认 externalize `node_modules` 中的第三方依赖。
+
+当 environment 是 DOM 或类浏览器自定义 environment 时，使用 `transformMode: 'browser'`。在这个模式下，如果没有自定义 `resolve.conditionNames`，Rstest 会添加 `browser` condition 来解析面向浏览器的 package entry，并默认 bundle 第三方依赖，从而让 browser field 替换和被打包的资源生效。
```
fi3ework: IIUC, is the 'browser' here actually the intersection of 'jsdom' and 'happy-dom' with their respective specialized logic removed? And is 'node' the 'node' from testEnvironment?

9aoy: You are absolutely right!

fi3ework: i'm a bit concerned, if that's the case, then `transformMode` essentially represents an environment, and its actual function isn't limited to just "transform." the naming feels a bit strange to me. 

what if `environment`(not using `testEnvironment` since no `browser` enum for in the existing `testEnvironment` config now).

9aoy: I’m a bit worried that `environment` may be confusing here, since we already have `testEnvironment` as the config field and the custom environment object itself.

What do you think about `buildTarget` instead?

It better reflects what this option actually controls: the build/resolve defaults used for the test code, such as browser-oriented resolution and dependency bundling, rather than the runtime environment object itself.

fi3ework: what about target, just like https://rsbuild.rs/config/output/target#outputtarget

9aoy: `target` sounds good to me. 

I’d prefer to keep it scoped under `testEnvironment`:

```ts
testEnvironment: {
  name: './custom-env.mjs',
  target: 'web',
}
```

9aoy:
@codex review

9aoy:
@codex review

9aoy:
@codex review

9aoy:
@codex review

9aoy:
@codex review

PR #1289, opened 2026-05-20
feat(core): experimental softMode — reuse workers with per-file env reset

pkasarda:
**Draft — opening for visibility, not for review yet.** No CC.

## What

Adds `experiments.softMode: boolean | { maxFilesPerWorker?: number }`
to `RstestConfig`. When enabled, the worker pool reuses processes
across test files (like `isolate: false`) but also resets per-file env
state between them — DOM, storage, fake timers, spies, prototype
mutations — so most tests written against Jest's per-file-context
isolation still pass without code changes.

```ts
defineConfig({
  experiments: { softMode: true },
});

// or with tuning:
defineConfig({
  experiments: { softMode: { maxFilesPerWorker: 10 } },
});
```

## Why

In our private workspace ~70-file jsdom-heavy libs were spending the
bulk of wall time on per-file `fork()` + `new JSDOM()` + setupFiles.
With `isolate: true` (default), each file pays that fixed cost. With
`isolate: false`, the workers are reused but ANY cross-file state
(DOM body, prototype mutations from user-event/react-aria, fake timers,
spies) leaks — practically unusable.

`experiments.softMode` is the in-between: reuse the worker AND
mechanically clean the per-file-isolated state the worker level can
see.

## What's reset between files

- `document.body` / `document.head` innerHTML, `window.history`, scroll
- `localStorage`, `sessionStorage`, `document.cookie` (path=/ only)
- DOM prototype descriptors on `HTMLElement`, `Element`, `Node`
  (snapshotted at env-setup, restored before next file). Also drops
  own-keys added since snapshot (e.g. vendor-installed
  `Symbol('patched-focus')` markers from user-event's `patchFocus`).
- Fake timers (`useRealTimers()` on the prior file's `Rstest` api)
- Tinyspy spies via worker-scope `restoreAll()`
- Pending async from the prior file is drained + absorbed at the top
  of `preparePool` so leftover `unhandledRejection`s from the prior
  file don't misattribute to the next file's slot

## Heap pressure → worker recycling

`--trace` profiling revealed individual tests running **2-3× slower**
under softMode vs `isolate: true` on a 74-file jsdom-heavy lib. Heap
instrumentation showed why:

| File seq in worker | heapUsed | heapTotal | rss |
|---:|---:|---:|---:|
| 1 | 71 MB | 136 MB | 254 MB |
| 10 | 745 MB | 937 MB | 1.5 GB |
| 50 | ~3.5 GB | ~3.7 GB | ~4 GB |
| 65 | 4.0 GB | 4.1 GB | 4.5 GB |

Sharing a JS heap across many React-heavy files = monotonic growth
(vendor module instances, React fiber roots, JSDOM nodes, accumulated
closures). By file 50 the worker GC-thrashes at 4 GB. This is the
hard limit of soft-mode reuse without `vm.Context` isolation.

**Fix**: `PoolRunner` now has `maxTasks` — when a soft-mode worker
hits N tasks, `isUsable()` flips false and the pool's existing dispose
path spawns a fresh worker. Heap pressure can't accumulate past a
known ceiling. Default cap **20** for `isolate !== true`; strict
isolate is single-use and unaffected. The cap is user-configurable
via `experiments.softMode: { maxFilesPerWorker: N }` — dial it up for
light-test libs, down for heavier ones. With recycle at 20, the
worst lib's reports-module went from individual tests being 2-3×
slower to roughly parity with `isolate: true`, while still amortizing
fork + jsdom-init across the 20-file window.

## What doesn't change

- `isolate: true` semantics — unchanged.
- `isolate: false` semantics — unchanged (still no per-file reset).
- Public type `RstestConfig.isolate` stays `boolean`.

## Why `experiments.*`

The reset semantics are pragmatic (curated to what we found in
practice). I'd rather mark this as experimental and iterate than
freeze a contract that turns out incomplete. A `logger.warn` fires if
a user sets both `isolate: false` AND `experiments.softMode` enabled.

## Files

- `packages/core/src/types/config.ts` — public `experiments?: { softMode?: boolean | SoftModeOptions }`, new `SoftModeOptions` interface, internal `NormalizedConfig.isolate: boolean | 'soft'`
- `packages/core/src/config.ts` — normalize any truthy `experiments.softMode` to internal `isolate: 'soft'`, warn on `isolate: false` + softMode conflict
- `packages/core/src/pool/index.ts` — extract `maxFilesPerWorker` from `experiments.softMode` and thread into `PoolOptions`
- `packages/core/src/pool/pool.ts`, `pool/types.ts` — reuse runners for `'soft'` in addition to `false`; pass `maxTasks` to `PoolRunner` (default 20, override via config)
- `packages/core/src/pool/poolRunner.ts` — `tasksRun` counter, `recordTaskCompleted()`, `isUsable()` cap
- `packages/core/src/runtime/worker/runInPool.ts` — env caching, prototype snapshot/restore, fake-timer reset, tinyspy `restoreAll`, drain-pending-async absorber, `beforeExit` teardown, optional `RSTEST_HEAP_TRACE=1` diagnostic
- `e2e/soft-mode/` — jsdom fixture covering DOM reset, prototype restore, fake-timer reinstall, spy restore, and absorber path (3 files: `a-init` / `b-leak` / `c-verify`)
- `e2e/soft-mode-reuse/` — `maxWorkers: 2` × 4 files, asserts actual worker reuse via pid log
- `e2e/soft-mode-happy-dom/` — happy-dom env path
- `e2e/soft-mode-recycle/` — `maxFilesPerWorker: 1` × 4 files, asserts unique pid per file (cap honored end-to-end)

## Testing

- `pnpm --filter @rstest/core test` — 242/242
- `e2e --isolate false` — 475/475 (was 472 pre-changes, +3 new soft-mode fixtures)
- `e2e` default — 627/629 (2 pre-existing browser-mode failures unrelated)

Validated in a private 348-file workspace: workspace-mode green at
3348/3348. Bench on the worst-case lib (74-file reports-module):
softMode + recycle median 23 s wall vs vanilla `isolate: true` median
33 s wall (**−30%**). Workspace-mode delta is smaller (~5%) because
parallelism across the default 11 workers dilutes the per-worker
amortization.

## Status

Marking draft. Not requesting review yet; just opening so I and a
colleague can see the cumulative diff in one place.

Followups before un-drafting:
- Heap-based recycle (recycle when heap > N MB) as an alternative to
  task-count
- Document expected wins/limits in the user-facing docs

### DIFF
--- a/e2e/soft-mode-happy-dom/fixtures/rstest.config.mts
+++ b/e2e/soft-mode-happy-dom/fixtures/rstest.config.mts
@@ -1,0 +1,14 @@
+import { defineConfig } from '@rstest/core';
+
+export default defineConfig({
+  experiments: {
+    softMode: true,
+  },
+  testEnvironment: 'happy-dom',
+  // Force single worker so file-a always precedes file-b — assertions
+  // about cross-file state reset are deterministic.
+  pool: {
+    maxWorkers: 1,
+  },
+  setupFiles: ['./test/setup.ts'],
+});
--- a/e2e/soft-mode-happy-dom/fixtures/test/a-init.test.ts
+++ b/e2e/soft-mode-happy-dom/fixtures/test/a-init.test.ts
@@ -1,0 +1,20 @@
+import { describe, expect, it } from '@rstest/core';
+
+declare global {
+  // eslint-disable-next-line no-var
+  var __soft_worker_pid__: number | undefined;
+}
+
+globalThis.__soft_worker_pid__ ??= process.pid;
+
+describe('soft mode happy-dom — file A', () => {
+  it('owns a fresh DOM body', () => {
+    expect(document.body.innerHTML).toBe('');
+    document.body.innerHTML = '<div id="from-a">x</div>';
+  });
+
+  it('mutates storage', () => {
+    expect(localStorage.length).toBe(0);
+    localStorage.setItem('from-a', 'leaked-if-soft-reset-broken');
+  });
+});
--- a/e2e/soft-mode-happy-dom/fixtures/test/b-verify.test.ts
+++ b/e2e/soft-mode-happy-dom/fixtures/test/b-verify.test.ts
@@ -1,0 +1,33 @@
+import { describe, expect, it } from '@rstest/core';
+
+declare global {
+  // eslint-disable-next-line no-var
+  var __soft_worker_pid__: number | undefined;
+}
+
+describe('soft mode happy-dom — file B', () => {
+  it('runs in the same worker as file A', () => {
+    expect(globalThis.__soft_worker_pid__).toBe(process.pid);
+  });
+
+  it('starts with a clean DOM body', () => {
+    expect(document.body.innerHTML).toBe('');
+    expect(document.querySelector('#from-a')).toBeNull();
+  });
+
+  it('starts with clean storage', () => {
+    expect(localStorage.length).toBe(0);
+    expect(localStorage.getItem('from-a')).toBeNull();
+  });
+
+  it('HTMLElement.prototype.focus descriptor restored across files', () => {
+    // setup.ts records the descriptor SHAPE on each module re-eval BEFORE
+    // re-patching. With softMode handling happy-dom, file-a saw 'value'
+    // (fresh DOM) and file-b also saw 'value' because softResetEnv's
+    // protoSnapshot restore wiped file-a's getter-only patch.
+    const history = (
+      globalThis as { __soft_focus_descriptor_history__?: Array<string> }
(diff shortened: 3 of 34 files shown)


pkasarda:
Superseded by smaller standalone PRs that don't bundle e-conomic-specific exhaust:

- #1291 — pool.memoryLimit heap recycle (the recycle bit, generalized)
- #1292 — --heapProfile diagnostic flag

The full softMode story turned out to need vendor-specific reset axes (DOM proto snapshot for user-event, drain-pending-async for msw v1) that aren't defensible as upstream defaults. Keeping those in a private fork; what's defensible upstream is just the building blocks above.

PR #1299, opened 2026-05-21
feat(core): make --reporters the canonical CLI flag

fi3ework:
## Summary

The CLI flag was named `--reporter` (singular), while the config field is `reporters` (plural). It also diverged from vitest's and jest's canonical `--reporters`, costing migration ergonomics. This PR makes `--reporters` the canonical CLI flag and keeps `--reporter` working as an alias.

- New canonical flag `--reporters <name>` on both `rstest [run|watch]` and `rstest merge-reports`. Repeatable for multiple reporters.
- `--reporter` continues to work via cac's native comma-separated alias (`--reporters, --reporter <name>`). cac routes both names to the canonical `options.reporters` key via mri, so consumer code reads a single field.
- No deprecation warning — the alias stays supported.
- `CommonOptions.reporters` typed as `string | string[]` to reflect cac's actual runtime shape (single occurrence → `string`, multiple → `string[]`); `castArray` already handles both.
- Docs (en + zh) updated to use `--reporters` throughout; the alias is mentioned once on the reporters guide.
- E2E tests and the `merge-reports` "missing blob directory" error hint switched to the canonical `--reporters` so they no longer recommend the legacy flag.

## Related Links

- vitest CLI config: https://github.com/vitest-dev/vitest/blob/main/packages/vitest/src/node/cli/cli-config.ts


fi3ework:
@9aoy if you're fine with the CLI renaming, we gonna merge it right before 0.10.2 release case the doc is updated alongside.

9aoy:
https://vitest.dev/config/reporters.html
image

image


It seems that vitest supports both `--reporter` and `--reporters`.

9aoy (approved):

PR #1235, opened 2026-05-09
feat(core): add threads pool implementation

fi3ework:
### Background

Rstest only ships a `forks` pool today. For test suites dominated by small, fast files, the `child_process.fork` spawn cost (~10x slower than `worker_threads`) dominates run time, and each worker carries its own V8 heap, which can OOM under high parallelism.

### Implementation

Host side:
- Add `ThreadsPoolWorker` backed by `node:worker_threads`; refactor lifecycle plumbing (emitter, stderr capture, exit state, stdio attach) shared with `ForksPoolWorker` into a new `BasePoolWorker`.
- `createPoolWorker` factory dispatches on `task.worker` (`'forks' | 'threads'`) with an exhaustive `_exhaustive: never` switch; plumb `pool.type` through `PoolTask` instead of the prior hardcoded `'forks'`.

Worker side (mirrors the host structure):
- `WorkerChannel` interface + `BaseChannel` abstract + `ForksChannel` / `ThreadsChannel` impls under `runtime/worker/channels/`. The factory in `channels/index.ts` auto-detects (`!isMainThread && parentPort !== null`) and reuses the same `PoolWorkerKind` union + exhaustive switch as the host factory.

Shared utilities:
- Extract `StderrCapture` (1MB tail buffer) and promote `toError` into `utils/helper.ts` for reuse across pool/runner/factory paths.

Tests + docs:
- `tests/pool/threadsPool.test.ts` (13 cases) and e2e fixture (`e2e/threads-pool/`); the fixture's `testWorker.mjs` auto-detects channel like the production worker.
- CLI hint and EN/ZH docs gain an independent "Choose a pool type" section with user-perspective trade-off bullets.

### User Impact

New `pool: 'threads'` opt-in for users with small / fast test files, memory-bound parallelism, or no per-file process-state isolation requirements. Default stays `forks`.

```
                       Host
            +------------------------+
            |   PoolRunner / Pool    |
            +-----------+------------+
                        | createPoolWorker(task.worker)
            +-----------+------------+
            |     BasePoolWorker     |  emitter, stderr, exit, attachStd*
            +-----+-------------+----+
                  |             |
         +--------+----+   +----+---------+
         | ForksPool   |   | ThreadsPool  |
         | Worker      |   | Worker       |
         | fork()      |   | new Worker() |
         +-----+-------+   +------+-------+
               | IPC              | postMessage
                       Worker
            +------------------------+
            |   runtime/worker entry |
            +-----------+------------+
[...]

PR #1321, opened 2026-05-26
ci(codspeed): split CPU per-PR / memory push-to-main, drop codspeed-macro

fi3ework:
## Summary

CodSpeed was disabled because the walltime + `codspeed-macro` setup exhausted the bare-metal runner quota. This PR brings per-commit perf tracking back without that dependency.

- **CPU bench** runs on every PR and push to main, using `mode: simulation` (Callgrind) on `ubuntu-latest`. The three bench fixtures pin `pool: 'threads'` because Callgrind runs with `--instr-atstart=no` — forked workers go through `fork+execve` and restart with instrumentation off, so their work would not be measured. Threads share the host process's Callgrind state, so the whole test pipeline is captured.
- **Memory bench** is demoted from per-PR to push-to-main only (plus manual dispatch). The jsdom + antd + three fixture is heavy and memory regressions are slow-moving — a per-commit baseline on main plus on-demand dispatch is a better fit than burning PR CI minutes for low-signal diffs. Note: CodSpeed's memory instrument uses memtrack (eBPF + libc malloc uprobes + `sched_process_fork` tracepoint), so forked-worker allocations are tracked across the fork boundary — visibility was never the issue.
- **Cleanup**: drop the weekly schedule and the `check-skip` job — both existed only to keep the macro-runner CPU baseline maintained on a low-cost cadence. Per-commit triggers cover baseline maintenance directly. Concurrency group now includes \`event_name\` so a manual dispatch on the same SHA does not collide with the push run.

## Related Links

- CodSpeed CPU instrument: https://codspeed.io/docs/instruments/cpu
- Threads pool feature: #1235

### DIFF
--- a/.github/workflows/codspeed.yml
+++ b/.github/workflows/codspeed.yml
@@ -7,9 +7,6 @@ on:
         description: 'Git ref to benchmark.'
         required: false
         type: string
-  schedule:
-    # Run once a week on main to build a continuous baseline for CPU benchmark
-    - cron: '0 0 * * 0'
   # Allow manual runs on any branch (CodSpeed backtest + on-demand profiling)
   workflow_dispatch:
     inputs:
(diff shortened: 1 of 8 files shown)


fi3ework:
close as even `threads` pool also include a large quantity of system call. and no pure JS heavy function in Rstest worth run on CodSpeed. will use walltime bench as before.

PR #1325, opened 2026-05-27
fix(cli): support filter args after coverage

kwonoj:
## Summary

Minor adjustment for cli handling, found 

```
rstest --coverage somefile.ts
```

omits file filter, looks like it is treated as input for the boolean flag for `coverage` instead.  If run 

```
rstest somefile.ts --coverage
```

file filter works fine.

PR tries to align flag behavior same as `browser`, having `coverage.enabled` support and allow to accept filters next to the coverage flag.

## Checklist


9aoy (approved):

9aoy:
Nice catch 👍

Issue #1332, opened 2026-05-28
[Feature]: Extending CLI override support scope (and how wide?)


kwonoj:
### What problem does this feature solve?

I found overriding current config via CLI flag have some discrepancy. For example, `--coverage.provider` is supported cli flags. However, `--coverage.reporters` are not. Peeking current implementation, probably some more flags are not fully overridable via cli flags.

This issue is generic, high level question regarding behavior

- Is it expected rstest supports overriding existing config via cli?
- If so, what are expected range of args to be expected?

(lastly, would it ok to extend coverage.provider as well?)

### What does the proposed API look like?

Not asking exact api changes yet, more of design questions for cli / config behaviors for now.

fi3ework:
1. Is it expected rstest supports overriding existing config via cli?

	for most configs, yes, it should.

2. If so, what are expected range of args to be expected?

	i cannot give you a definitive answer yet. however, we will likely support most options using primitive values unless there is absolutely no need for overrides in real-world scenarios.
	
	- we will support `coverage.reporters` in the next patch, as it is easy to implement.
	- does "extend `coverage.provider`" refer to a custom provider?

kwonoj:
> coverage.provider

It's mistake, I was referring coverage.reporters.

Can make pr on my end too, wanted to check if this is something acceptable before making any changes.

fi3ework:
@9aoy PTAL

IMHO, custom reporter providers are a feature worth introducing. if we decide to support them, the implementation should be straightforward. but we can try to design the API carefully so it can remain stable for version 1.0.0.

9aoy:
> [@9aoy](https://github.com/9aoy) PTAL
> 
> IMHO, custom reporter providers are a feature worth introducing. if we decide to support them, the implementation should be straightforward. but we can try to design the API carefully so it can remain stable for version 1.0.0.

Coverage reporters in Rstest follow the standard Istanbul reporter API. Therefore, instead of introducing a Rstest-specific coverage reporter API, we should document this clearly and make sure custom Istanbul-compatible coverage reporters can be loaded properly.

PR #1336, opened 2026-05-28
feat(core): support more coverage CLI options

9aoy:
## Summary

This PR expands coverage CLI overrides beyond `--coverage.reporters`, which is already supported on main, so users can configure common coverage options directly from the command line. It adds CLI support for coverage include/exclude patterns, reports directory, report-on-failure, clean, and allowExternal, with unit/e2e coverage and synchronized English/Chinese docs. The Chinese CLI options table is also reorganized alphabetically for easier scanning.

## Related Links

https://github.com/web-infra-dev/rstest/issues/1332

## Checklist


fi3ework:
@codex review

packages/core/src/cli/init.ts
```diff
@@ -214,12 +219,40 @@ function mergeWithCLIOptions(
...
         shouldEnableCoverage = true;
       }
+      if (options.coverage.include !== undefined) {
+        config.coverage.include = castArray(options.coverage.include);
+        shouldEnableCoverage = true;
+      }
+      if (options.coverage.exclude !== undefined) {
+        config.coverage.exclude = castArray(options.coverage.exclude);
+        shouldEnableCoverage = true;
+      }
```
fi3ework, line 232: This replaces the default `exclude` list entirely instead of extending it. After `rstest run --coverage --coverage.exclude='src/generated/**'`, the resulting `exclude` becomes just `['src/generated/**']`, so `**/node_modules/**`, `**/*.d.ts`, `**/*.{test,spec}.[jt]s` etc. all stop being excluded and start showing up in the report.

The `coverage.exclude` docs (unchanged here) still say *"Any patterns specified here will be merged with default values."*, which now contradicts the CLI behavior.

Could we concatenate with the existing `config.coverage.exclude` instead, and add an e2e case asserting a `node_modules/...` file stays excluded when `--coverage.exclude` is passed?

9aoy: Good catch, thanks! Fixed in `bd5d377c`: `--coverage.exclude` now appends CLI patterns to the existing `config.coverage.exclude` instead of replacing it. I also updated the unit assertion and added an e2e regression that creates a `node_modules/ignored-package/index.ts` file and verifies it stays out of the coverage text report when `--coverage.exclude` is passed.

Validation:
- `pnpm rstest packages/core/tests/cli/init.test.ts`
- `pnpm --filter @rstest/core build`
- `cd e2e && pnpm test test-coverage-v8/index.test.ts`
- `pnpm run check-unused`

Issue #1294, opened 2026-05-21
API stabilization of 1.0.0


fi3ework:
Tracking list of open API design decisions to settle before the 1.0 release. Items here would be breaking (or potentially breaking) to change after 1.0; deferred / will-add-later / unimplemented items go into separate roadmap issues.

Each item is numbered ([N]) so it can be referenced in discussion (e.g. "let's talk about [5]").

**Convention**: when you check a box, fill the 3-field template under it so the rationale is captured inline.

```
- [x] [N] item
  - Picked: <chosen option>
  - Why: <one-line reason>
  - Refs: <PR / ADR / discussion link>
```

---

## ⚪ Not yet reviewed (4)

- [ ] [21] File-level hook naming: `onTestFileStart` / `onTestFileReady` / `onTestFileResult` or `onTestModuleStart` / `onTestModuleEnd`
  - Picked: 
  - Why: 
  - Refs: 

- [ ] [22] Reporter data shape: plain object or wrapper classes (`TestModule` / `TestSuite` / `TestCase`), or mark the whole `Reporter` interface as Best-effort tier and defer
  - Picked: 
  - Why: 
  - Refs: 

- [ ] [23] `TestResultStatus` literal lock-in: commit current values `'pass' | 'fail' | 'skip' | 'todo'` as the Stable contract for custom reporters, or keep room to rename
  - Picked: 
  - Why: 
  - Refs: 

- [ ] [26] `browser.browser` field shape: single-browser field or `browser.instances[]` array
  - Picked: 
  - Why: 
  - Refs: 

## ✅ Decided (32)

- [x] [13] `shard` config object form `{ count, index }` tier: keep as a Stable config-side form alongside the CLI flag, or limit to CLI flag only
  - Picked: Option 2 — CLI-only. Drop the `shard` config field; expose sharding only via the `--shard <index>/<count>` flag.
  - Why: shard is per-invocation — every CI runner must pass a *different* index, which a value committed to the shared config file cannot express. Mirrors jest, which carries shard on `Argv`/`GlobalConfig` and never adds it to user config types. Internally the resolved shard now rides on an internal `ResolvedRstestConfig` / `NormalizedConfig` that the CLI injects; the obsolete per-project shard validation is removed; docs move out of `config/test/` into the CLI guide (`rstest run` -> Run a shard).
  - Refs: #1449

- [x] [7] `pool` shape and whether to remove `minWorkers` before 1.0.0
[...]

PR #1364, opened 2026-06-03
fix(core): type runtime hook registration as sync

nnnnoel:
## Summary

- Narrow the return type of the `beforeAll`, `afterAll`, `beforeEach`, and `afterEach` registration APIs to `void`. The previous `MaybePromise<void>` return type could make type-aware lint rules such as `@typescript-eslint/no-floating-promises` report normal hook registration calls as unhandled promises, even though registration is synchronous.
- Keep the hook callback listener types as `MaybePromise` so async hooks and returned cleanup functions continue to work as before.
- Align the registration return type with Vitest's hook API typing, which also exposes these hook registration functions as returning `void`.
- This fits the 1.0 API stabilization work because it tightens the public hook registration contract before the API is locked.

## Related Links

- https://github.com/web-infra-dev/rstest/issues/1294
- https://typescript-eslint.io/rules/no-floating-promises/
- https://github.com/vitest-dev/vitest/blob/main/packages/runner/src/hooks.ts#L154

## Checklist

### DIFF
--- a/packages/core/src/runtime/runner/runtime.ts
+++ b/packages/core/src/runtime/runner/runtime.ts
@@ -91,7 +91,7 @@ class RunnerRuntime {
   afterAll(
     fn: AfterAllListener,
     timeout: number = this.runtimeConfig.hookTimeout,
-  ): MaybePromise<void> {
+  ): void {
     const currentSuite = this.getCurrentSuite();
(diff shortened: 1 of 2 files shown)


fi3ework (approved):

fi3ework:
thank you!

PR #1368, opened 2026-06-03
fix(core): prevent late console logs from failing the run (isolate: false)

### LINKED ISSUE #1367: [Bug]: "[birpc] rpc is closed" when a console call runs after a test file finishes (isolate: false)
### Version

```sh
System:
  OS: macOS 15.0.1
  CPU: (14) arm64 Apple M3 Max
Binaries:
  Node: 22.22.3
npmPackages:
  @rstest/core: ^0.10.3 => 0.10.3
```

### Details

With `isolate: false`, all test files share a single worker. rstest gives each file its own birpc channel and `$close()`s the previous one when the next file starts (`preparePool` in `dist/worker.js`). The worker's console interceptor forwards every `console.*` write over that channel:

`createCustomConsole` (`dist/0~console.js`) → `onConsoleLog` → `emitInterceptedLog` (`dist/worker.js:368`) → `await rpc.onConsoleLog(log)`.

That call is fire-and-forget (the interceptor never awaits it) and has no guard against a closed channel. So when any code writes to a **captured** `console` reference from an async callback that outlives its test file — a logger that flushes from a `setTimeout`/microtask, a compiler `done` hook, etc. — the write hits the already-closed channel and birpc throws:

```
[birpc] rpc is closed, cannot call "onConsoleLog"
```

as an `unhandledRejection`. Two consequences:

1. **It fails the run** even though every assertion passed.
2. **It blames the wrong file.** The rejection is caught by whichever file's `process.on('unhandledRejection')` handler is currently registered, so the failure is attributed to the file that happens to be running when the late log fires — not the file that produced it.

In the minimal repro the log comes from `a.test.ts:7`, but `b.test.ts` is marked as failed. rstest even reports `candidateFiles: a.test.ts:7`, confirming the source ≠ the failed file.

Stack:

```
unhandledRejection: [birpc] rpc is closed, cannot call "onConsoleLog"
    at _call (@rstest/core/dist/8549.js:77:28)
    at Proxy.sendCall (@rstest/core/dist/8549.js:152:41)
    at emitInterceptedLog (@rstest/core/dist/worker.js:368:23)
    at emitLog (@rstest/core/dist/worker.js:183:9)
    at Object.onConsoleLog ...
```

This is easy to hit in real suites because loggers commonly capture `console` once and write later. With `isolate: true` each file has its own worker that is torn down together with the file, so the late write never reaches a closed-but-shared channel and the problem does not appear.

Possible directions (whichever fits rstest's design): guard `emitInterceptedLog` against a closed channel (an `isClosed` check / `try-catch`), drain pending console calls before `$close()`, or drop/redirect console writes that arrive after the channel closes instead of throwing.

Minimal repro files (also in the repo linked below):

```ts
// rstest.config.ts
import { defineConfig } from '@rstest/core'

export default defineConfig({
  isolate: false,
  pool: { maxWorkers: 1 }, // force A and B onto the same shared worker
})
```

```ts
// a.test.ts
import { test } from '@rstest/core'

const log = globalThis.console // captured reference, like a logger

test('schedules a console call that outlives the test file', () => {
  setTimeout(() => {
    log.log('late log from a.test.ts')
  }, 200) // not awaited -> runs after this file's tests finish
})
```

```ts
// b.test.ts
import { test } from '@rstest/core'

test('keeps the worker alive', async () => {
  await new Promise((resolve) => setTimeout(resolve, 1000))
})
```

### Reproduce link

https://github.com/fi3ework/rstest-rpc-closed-repro

### Reproduce Steps

1. `npm install`
2. `npm test` (or `npx rstest run`)
3. The run fails: `b.test.ts` reports `unhandledRejection: [birpc] rpc is closed, cannot call "onConsoleLog"`, while the actual source is the `setTimeout` callback in `a.test.ts`. All tests themselves pass.

fi3ework:
## Summary

With `isolate: false`, all test files share one worker but each file owns its own birpc channel. The host disposes a file's channel as soon as the file finishes, while the worker only closes it when the next file starts. A console reference captured by a file (e.g. a logger that flushes from a late `setTimeout`/microtask) could fire in or after that window and forward the log over a disposed/closed channel.

The console interceptor does not await the forward, so the rejected send surfaced as an `unhandledRejection` (`[birpc] rpc is closed`) that **failed the run even though every assertion passed**, and **blamed the wrong file** — whichever happened to be running when the late log fired.

Console forwarding is best-effort, with two failure modes now handled at their own layer:

- **Delivery failures** (a disposed/closed channel, a teardown race) are an expected lifecycle artifact. The worker forwards the log fire-and-forget and swallows the rejected send, so the late log is dropped instead of crashing the run — matching `isolate: true` where such logs are lost as the worker is torn down.
- **Handler failures** (a user `onConsoleLog` filter or a reporter's `onUserConsoleLog` throwing) would otherwise be hidden by that swallow. The host now surfaces them where they occur, without failing the otherwise-passing run.

The change lives in the pool-agnostic worker/host paths, so it applies to both the `forks` and `threads` pools.

Originally surfaced while migrating Rsbuild's e2e runner to rstest (web-infra-dev/rsbuild#7776), then reduced to the minimal repro in #1367.

## Related Links

- Closes #1367
- Originally found in web-infra-dev/rsbuild#7776 — https://github.com/web-infra-dev/rsbuild/pull/7776

## Checklist

### DIFF
--- a/e2e/log/console-hook-error.test.ts
+++ b/e2e/log/console-hook-error.test.ts
@@ -1,0 +1,36 @@
+import { dirname, join } from 'node:path';
+import { fileURLToPath } from 'node:url';
+import { describe, expect, it } from '@rstest/core';
+import { runRstestCli } from '../scripts';
+
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = dirname(__filename);
+
+// Companion to `late-console.test.ts` for https://github.com/web-infra-dev/rstest/issues/1367.
+// The worker forwards console output fire-and-forget (it never awaits the
+// result, and swallows the rejected send), so a user `onConsoleLog` filter (or a
+// reporter's `onUserConsoleLog`) that throws while the host handles a forwarded
+// log cannot travel back to fail the test — and would be hidden by that swallow.
+// The host must surface that error as a diagnostic instead, while the passing
+// test stays green. Assert this for both pools (the host handler is pool-agnostic).
+describe('console hook error (issue #1367)', () => {
+  for (const pool of ['forks', 'threads'] as const) {
+    it(`should surface a throwing onConsoleLog hook without failing the run (${pool} pool)`, async () => {
+      const { cli, expectExecSuccess } = await runRstestCli({
+        command: 'rstest',
+        args: ['run', '--pool', pool],
+        options: {
+          nodeOptions: {
+            cwd: join(__dirname, 'fixtures/console-hook-error'),
+          },
+        },
+      });
+
+      await expectExecSuccess();
+
+      // The hook error must be surfaced on the host, not silently swallowed.
+      expect(cli.log).toContain('Failed to handle console log');
+      expect(cli.log).toContain('onConsoleLog hook boom');
+    });
+  }
+});
--- a/e2e/log/fixtures/console-hook-error/a.test.ts
+++ b/e2e/log/fixtures/console-hook-error/a.test.ts
@@ -1,0 +1,6 @@
+import { expect, test } from '@rstest/core';
+
+test('passes even though the onConsoleLog hook throws', () => {
+  console.log('hello from a');
+  expect(1).toBe(1);
+});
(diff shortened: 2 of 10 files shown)


fi3ework:
@codex review

PR #1376, opened 2026-06-04
fix(core): share imported module state and keep context-bound APIs live across files under isolate: false

### LINKED ISSUE #1373: [Bug]: Imported module state is not shared across test files under `isolate: false`
## Version

```sh
System:
  OS: macOS (Darwin 24)
npmPackages:
  @rstest/core: 0.10.3
```

Also present on `main` (commit `956d5cc2`).

## Details

Under `isolate: false`, a module that is imported into multiple test files is **re-evaluated once per test file** within the same worker, instead of once per worker. Module-level state therefore does not persist across files, even though the worker process is reused.

**Expected:** with `isolate: false`, a module evaluated in a worker should be reused across the test files that run in that worker (evaluate once); only the test-entry module should re-run per file. State held at module scope (singletons, lazily-initialized resources, caches) should survive across files.

**Actual:** the module's top-level code runs again for every file, resetting all module-scope state.

### Root cause (verified in source)

Worker teardown unconditionally clears the entire module cache after every file when `!isolate`:

- `runtime/worker/runInPool.ts:419-423` — teardown calls `clearModuleCache()` whenever `!isolate`.
- `runtime/worker/loadEsModule.ts:349-352` — `clearModuleCache()` → `esmCache.clear()` (ESM / `.mjs` output). The CJS path is the same: `runtime/worker/loadModule.ts:336-339` → `moduleCache.clear()`.

The bundle is already structured to share state across files: every test entry statically imports a single per-worker runtime chunk (`rstest-runtime.mjs`) that owns the only module-instance cache (`__webpack_module_cache__`). Because that runtime chunk also lives in `esmCache`, `clearModuleCache()` evicts it too — so the next file re-instantiates the runtime, gets a fresh `__webpack_module_cache__`, and every bundled module factory (including imported helpers) re-executes.

The modules that *should* re-run per file already have a precise, targeted eviction path that does not require the blanket clear:

- `runtime/worker/runInPool.ts:329-332` injects `__rstest_clean_core_cache__()` per file.
- `core/plugins/moduleCacheControl.ts:15-26` — it deletes only `__webpack_module_cache__['@rstest/core']` and the registered setup-file ids (`setupIds`).

So `@rstest/core` and setup files are already reset per file by their own mechanism; the additional blanket `clearModuleCache()` is what defeats cross-file sharing of every other module.

### Suggested direction

Make module-cache invalidation **selective** under `isolate: false`: keep the runtime chunk and already-evaluated non-entry modules in the cache, and evict only the current test-entry module (so its body re-runs), letting the existing `__rstest_clean_core_cache__` mechanism continue to handle `@rstest/core` + setup files. Note the test entry's `SourceTextModule` still needs to be evicted from `esmCache` for the entry to re-run.

## Reproduce

No external repro needed — the behavior is deterministic with a single worker (both files land in the same worker, so `shared.ts` should evaluate exactly once).

<details>
<summary>Minimal reproduction (key files)</summary>

```ts
// shared.ts — module-scope state; should initialize once per worker
console.log('[shared] evaluated, pid =', process.pid);
let value: number | undefined;
export const getValue = () => (value ??= Date.now());
```

```ts
// a.test.ts
import { getValue } from './shared';
test('a', () => {
  console.log('[a] getValue =', getValue());
  expect(getValue()).toBeDefined();
});
```

```ts
// b.test.ts
import { getValue } from './shared';
test('b', () => {
  console.log('[b] getValue =', getValue());
  expect(getValue()).toBeDefined();
});
```

```ts
// rstest.config.ts

export default defineConfig({
  isolate: false,
  // force both files into one worker for a deterministic repro
  pool: { maxWorkers: 1 },
});
```

</details>

## Reproduce Steps

1. Run `rstest`.
2. Observe that `[shared] evaluated` is logged **twice** (once per test file) with the **same** `pid`, and that `[a] getValue` / `[b] getValue` print **different** values.
3. Expected with `isolate: false`: `[shared] evaluated` logged **once**, and both files observe the **same** `getValue()`.

---

Encountered while migrating the e2e test suite in web-infra-dev/rsbuild#7776.

fi3ework:
## Summary

Under `isolate: false`, one worker process runs many test files: `@rstest/core` is reset per file (its module cache is evicted and re-imported) while user modules persist — a helper imported by several files is evaluated only once per worker. Two problems followed from that asymmetry, both fixed here.

### 1. Imported module state was not shared across files (#1373)

The worker teardown cleared the entire module cache after every file, evicting the shared per-worker runtime chunk that owns the only `__webpack_module_cache__`. The next file re-instantiated that runtime, so every bundled module factory ran again and module-scope state (singletons, caches, lazily-initialized resources) did not survive across files — defeating the main reason to opt into non-isolated mode.

Teardown invalidation is now selective:

- **Keep the shared runtime chunk** so `__webpack_module_cache__` — and the state of every already-evaluated non-entry module — persists across files. Only the test-entry and setup modules are evicted so their bodies re-run per file.
- **Accumulate per-entry asset maps** into one persistent per-worker map, so a later file's chunks stay resolvable by the kept runtime chunk's `import.meta` hooks (wasm / dynamic-import resolution). The map is reset only on a full cache clear.
- **Normalize setup paths with `pathe` before matching**, so setup files reliably re-run per file on Windows / mixed separators (previously a raw-path collection clobber left the targeted setup-reset list empty).

### 2. Context-bound APIs went stale when value-copied into a shared helper

Because user modules persist, any context-bound `@rstest/core` value captured at module top-level in a persisted helper — `export const test = base.extend({})`, a re-exported `onTestFinished`, `{ ...rstest }`, or a snapshotted `expect.poll` — froze to the **first** file's torn-down context and silently misbehaved from the second file on.

Fixed with a single live-binding contract: each context-bound member is a **stable forwarder, built once**, that re-resolves the running file's state from a per-file global slot at call time — never closing over a per-file instance. The collection surface (`runtimeAPI`) and the full runner surface (`runnerAPI`) are built once; `createRuntimeAPI` only constructs the file's runtime and publishes its slot. `expect` additionally self-delegates to the live `GLOBAL_EXPECT` to rescue an already-copied `expect.poll`/`.soft`. The contract — the four slots, the shared value-copy residual, and why `expect` is the one self-delegating member — is documented in one place (`runtime/api/index.ts`).

### 3. Shared module state went stale across watch rebuilds

Sharing user modules means keeping the runtime chunk (which owns the only `__webpack_module_cache__`) across files. In watch mode the worker pool is reused across rebuilds, so that kept cache also survived rebuild boundaries — under `isolate: false` a changed shared module was re-run with the **previous build's** exports until the process restarted.

A per-compile `buildId` is now threaded to the worker. When it changes (every watch rebuild), the worker fully flushes its module cache before loading the new build, so each rebuild re-evaluates shared modules. Intra-build sharing across files is unchanged.

### Behavior

With `isolate: false`, an imported module now evaluates once per worker and is shared across the files that run in it, and context-bound APIs always act on the running file; the test entry, setup files, and `@rstest/core` still reset per file. Applies to both ESM and CommonJS output.

A new e2e surface guard drives the whole context-bound API surface (`test.extend`, hooks, `describe`, `expect.poll`, `rstest.fn`/`clearMocks`) through a persisted shared helper from a non-first file; a watch regression test asserts that an edited shared module is re-evaluated on rerun; and the `isolate` docs note that module-scope code runs once per worker (use setup files for per-file setup).

## Related Links

- Fixes #1373

## Checklist

### DIFF
--- a/e2e/no-isolate/fixtures/sharing/a.test.ts
+++ b/e2e/no-isolate/fixtures/sharing/a.test.ts
@@ -1,0 +1,17 @@
+import { expect, test } from '@rstest/core';
+import { getSharedEvalId } from './shared';
+
+test('a: statically imported module is shared across files', () => {
+  expect(getSharedEvalId()).toBe(1);
+});
+
+test('a: dynamically imported module is shared across files', async () => {
+  const { getDynEvalId } = await import('./dynShared');
+  expect(getDynEvalId()).toBe(1);
+});
+
+test('a: setup re-ran for this file', () => {
+  expect((globalThis as Record<string, any>).__rstestSetupFor).toContain(
+    'a.test',
+  );
+});
--- a/e2e/no-isolate/fixtures/sharing/b.test.ts
+++ b/e2e/no-isolate/fixtures/sharing/b.test.ts
@@ -1,0 +1,17 @@
+import { expect, test } from '@rstest/core';
+import { getSharedEvalId } from './shared';
+
+test('b: statically imported module is shared across files', () => {
+  expect(getSharedEvalId()).toBe(1);
+});
+
+test('b: dynamically imported module is shared across files', async () => {
+  const { getDynEvalId } = await import('./dynShared');
+  expect(getDynEvalId()).toBe(1);
+});
+
+test('b: setup re-ran for this file', () => {
+  expect((globalThis as Record<string, any>).__rstestSetupFor).toContain(
+    'b.test',
+  );
+});
--- a/e2e/no-isolate/fixtures/sharing/dynShared.ts
+++ b/e2e/no-isolate/fixtures/sharing/dynShared.ts
@@ -1,0 +1,7 @@
+// Same contract as `shared.ts`, but reached via dynamic `import()` so the
+// dynamic-import + asset-resolution path is covered too.
+const g = globalThis as Record<string, any>;
+g.__rstestDynEvalCount = (g.__rstestDynEvalCount ?? 0) + 1;
+const evalIdAtLoad: number = g.__rstestDynEvalCount;
+
+export const getDynEvalId = (): number => evalIdAtLoad;
--- a/e2e/no-isolate/fixtures/sharing/mockShareFirst.test.ts
+++ b/e2e/no-isolate/fixtures/sharing/mockShareFirst.test.ts
@@ -1,0 +1,13 @@
+import { expect, test } from '@rstest/core';
+import { sharedMock } from './sharedMock';
+
+// Runs BEFORE mockShareSecond (alphabetical order, single worker). Leaves call
+// history on the shared-module mock so the next file can prove `clearMocks`
+// reset it across the file boundary.
+test('mockShareFirst: shared-module mock records calls', () => {
+  sharedMock(1);
+  sharedMock(2);
+  expect(sharedMock.mock.calls.length).toBe(2);
+  (globalThis as Record<string, any>).__mockShareFirstCalls =
+    sharedMock.mock.calls.length;
+});
--- a/e2e/no-isolate/fixtures/sharing/mockShareSecond.test.ts
+++ b/e2e/no-isolate/fixtures/sharing/mockShareSecond.test.ts
@@ -1,0 +1,13 @@
+import { expect, test } from '@rstest/core';
+import { sharedMock } from './sharedMock';
+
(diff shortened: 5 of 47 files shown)


fi3ework:
@codex review

fi3ework:
@codex review

9aoy (approved):

9aoy:
👍

PR #1384, opened 2026-06-04
refactor(core, adapters): single-source adapter cache-key & target→env logic

fi3ework:
## Summary

The three config adapters (`@rstest/adapter-rsbuild`, `@rstest/adapter-rslib`, `@rstest/adapter-rspack`) had each grown their own copy of the same config-mapping logic, and the copies had silently drifted:

- **Cache-key resolution** — resolving `buildDependencies` entries to normalized paths. rslib skipped `normalize()`, so the same dependency could yield a different cache key depending on which adapter ran.
- **`target` → `testEnvironment` mapping** — deciding whether a build target runs in Node, duplicated three ways.

This PR moves the shared logic into a single source, `packages/core/src/adapter.ts`, exposed via `@rstest/core/internal/adapter`:

- `resolveCacheDependency` — per-entry path resolution (absolute → normalize; relative → resolve against the config dir or root, then normalize).
- `resolveBuildCache` — the full `performance.buildCache` mapping shared by the rsbuild and rslib adapters (the rspack adapter keeps its own persistent-cache variant).
- `isNodeTarget` / `resolveTestEnvironmentFromTarget` — one definition of which targets count as Node, across single-string and array (rspack multi-target) shapes.

**No new runtime coupling.** The adapters previously imported `@rstest/core` only as types (erased at build time). To preserve that, each adapter's `rslib.config.mts` sets `externals: { '@rstest/core/internal/adapter': false }`, so these pure helpers are **bundled into the adapter's own dist** instead of left as a runtime import of a core subpath. The source stays single-sourced in core, but the shipped adapter carries its own copy — so a published adapter needs no particular core version to resolve the export. (This differs from `@rstest/browser`, which consumes its `internal/*` import at runtime and keeps it external.)

The rslib adapter intentionally keeps its own inverse target→env rule (libraries are Node-first, so an absent target defaults to `node` rather than `happy-dom`); only the shared "which targets are Node" predicate is reused.

## Checklist

### DIFF
--- a/packages/adapter-rsbuild/rslib.config.mts
+++ b/packages/adapter-rsbuild/rslib.config.mts
@@ -4,6 +4,14 @@ import { rsdoctorCIPlugin } from '../../scripts/rsdoctorPlugin';
 
 export default defineConfig({
   plugins: publishCheckPlugins(),
+  output: {
+    // Bundle the shared config helpers into the adapter's dist so it carries no
+    // runtime dependency on a core subpath; `false` opts this exact subpath out
+    // of autoExternal.
+    externals: {
+      '@rstest/core/internal/adapter': false,
+    },
+  },
   lib: [
     {
       format: 'esm',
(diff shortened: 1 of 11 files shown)


fi3ework:
@codex review

PR #1407, opened 2026-06-12
feat(core): add Module Federation compatibility mode

ScriptedAlchemy:
## Summary

Rstest's Node test runtime overrides `import()` and webpack chunk handlers for mocking, which breaks tests that load Module Federation remotes. This PR adds an opt-in federation compatibility mode so federated apps can be tested in Node:

- New `federation: boolean` config option and `--federation` CLI flag.
- Runtime shims in `mockRuntimeCode.js`, gated on `globalThis.__rstest_federation__`: a placeholder chunk-handler bootstrap, an `__rstest_dynamic_import__` fallback for non-mocked specifiers, and a `__webpack_require__.f` Proxy guard so federation's chunk loaders coexist with the mock runtime.
- Federation example apps (`examples/federation`), 8 unit tests, an e2e suite with 3 fixtures, and bilingual docs.

This mode pairs with the `@module-federation/rstest` Rsbuild plugin (module-federation/core#4320); the examples pin its latest pkg.pr.new canary (`1b77b1a`). Browser mode is unaffected. Supersedes #842.

Validation: lint, typecheck, build, unit tests, e2e, and federation examples all pass locally (webkit e2e skipped: missing system deps).

## Related Links

- https://github.com/module-federation/core/pull/4320
- https://github.com/web-infra-dev/rstest/pull/842

## Checklist


packages/core/src/core/plugins/mockRuntimeCode.js
```diff
@@ -1,5 +1,26 @@
...
+if (globalThis.__rstest_federation__) {
+  globalThis.__rstest_dynamic_import__ =
+    globalThis.__rstest_dynamic_import__ ||
+    function (specifier, importAttributes) {
+      return import(specifier, importAttributes);
+    };
+  // Mirror onto the local binding: bundler output in this chunk's scope calls
+  // `__rstest_dynamic_import__` as a free identifier, which resolves to the
+  // `var` above (undefined when this chunk itself was evaluated via vm/eval).
+  __rstest_dynamic_import__ = globalThis.__rstest_dynamic_import__;
```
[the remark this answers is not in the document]
ScriptedAlchemy: Confirmed — the runtime module is emitted inside its own IIFE, so the local `var` binding can never be observed by module factories (and SWC strips it from the emitted runtime anyway). Removed the dead local binding in 510f431f; the `globalThis` fallback is the only effective channel and is kept.

packages/core/src/runtime/worker/runInPool.ts
```diff
@@ -129,6 +129,11 @@ const preparePool = async (
   const taskContext = createNodeTaskContext();
   setRealTimers();
 
+  // `mockRuntimeCode.js` gates its Module Federation shims on this worker-wide
+  // flag, so it must be set before any bundle code is evaluated.
+  (globalThis as any).__rstest_federation__ =
+    context.runtimeConfig.federation === true;
```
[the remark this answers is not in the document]
ScriptedAlchemy: Good catch — fixed in 510f431f. `preparePool` now deletes `globalThis.__rstest_dynamic_import__` whenever `runtimeConfig.federation` is not enabled, so a reused worker (`isolate: false`) no longer leaks the shim from a prior federation-enabled file.

packages/core/src/core/plugins/mockRuntimeCode.js
```diff
@@ -42,6 +100,30 @@ __webpack_require__.rstest_original_module_factories = {};
 // not collide with `Object.prototype` members.
 __webpack_require__.rstest_mocked_ids_by_request = Object.create(null);
 
+//#region federation chunk handler placeholders
+// When `__webpack_require__.f` was populated before this runtime module ran,
+// the pre-seeding in the `set` trap above never saw it, and Module Federation
+// placeholder handlers that throw until the federation runtime initializes
+// may already be installed. Replace them with no-ops — the real runtime
+// overrides them once it boots.
+if (globalThis.__rstest_federation__ && __webpack_require__.f) {
```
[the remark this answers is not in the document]
ScriptedAlchemy: Fixed in 510f431f. After the placeholder cleanup, the pre-existing `f` object is now re-assigned through the `__webpack_require__` proxy, which routes it through the same `set`-trap wrapping — so `readFileVm` / `require` get the overwrite guard in this ordering too. Validated against the Module Federation example apps (main-app + component-app + node-local-remote) and the federation e2e suite.

packages/core/src/core/plugins/mockRuntimeCode.js
```diff
@@ -1,5 +1,23 @@
...
+// the function-argument injection that normally provides that hook, so the
+// free-identifier lookup falls through to `globalThis`. Provide a fallback
+// there that relies on Node's native dynamic import. (This runtime module is
+// emitted inside its own IIFE, so a local binding could never be observed by
+// module factories — the global is the only effective channel.)
+if (globalThis.__rstest_federation__) {
+  globalThis.__rstest_dynamic_import__ =
+    globalThis.__rstest_dynamic_import__ ||
+    function (specifier, importAttributes) {
+      return import(specifier, importAttributes);
```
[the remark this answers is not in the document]
ScriptedAlchemy: Valid — fixed in af97a666. The fallback now routes absolute filesystem specifiers (Windows drive letters and POSIX paths) through `pathToFileURL` before native `import()`, mirroring the normalization in `resolveDynamicImport.ts`. Added e2e coverage in the federation basic fixture that imports a module by raw absolute path through the fallback, which runs on the Windows CI matrix.

packages/core/src/core/plugins/mockRuntimeCode.js
```diff
@@ -1,5 +1,33 @@
...
+// evaluated via vm/eval — instead of the worker's `loadModule` — never receive
+// the function-argument injection that normally provides that hook, so the
+// free-identifier lookup falls through to `globalThis`. Provide a fallback
+// there that relies on Node's native dynamic import. (This runtime module is
+// emitted inside its own IIFE, so a local binding could never be observed by
+// module factories — the global is the only effective channel.)
+if (globalThis.__rstest_federation__) {
+  globalThis.__rstest_dynamic_import__ =
+    globalThis.__rstest_dynamic_import__ ||
+    function (specifier, importAttributes) {
```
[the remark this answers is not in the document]
ScriptedAlchemy: Valid — fixed in 773bf1f0. `preparePool` now drops `globalThis.__rstest_dynamic_import__` unconditionally before each file, so a reused worker (`isolate: false`) never carries the previous bundle's hook: non-federation files keep strict opt-in behavior, and federation files get a fresh hook reinstalled by the runtime module, bound to the current bundle's vm dynamic-import context.

packages/core/src/core/plugins/mockRuntimeCode.js
```diff
@@ -1,5 +1,33 @@
...
+// dynamic imports to `__rstest_dynamic_import__(<abs-path>)` calls. Chunks
+// evaluated via vm/eval — instead of the worker's `loadModule` — never receive
+// the function-argument injection that normally provides that hook, so the
+// free-identifier lookup falls through to `globalThis`. Provide a fallback
+// there that relies on Node's native dynamic import. (This runtime module is
+// emitted inside its own IIFE, so a local binding could never be observed by
+// module factories — the global is the only effective channel.)
+if (globalThis.__rstest_federation__) {
+  globalThis.__rstest_dynamic_import__ =
+    globalThis.__rstest_dynamic_import__ ||
```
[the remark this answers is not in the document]
ScriptedAlchemy: Looked into this one carefully — I don't think it's actionable, for three reasons:

1. The setup entries and the test entry share a single runtime chunk (`rstest-runtime`), which is evaluated once per file run (worker module caches). A "refresh per entry" cannot be expressed in this runtime module because it never re-executes for the test entry; clearing the global between setup and test evaluation in the worker would instead leave federation chunks with **no** hook, since the runtime won't run again to reinstall it.
2. The binding divergence between the setup and test vm contexts only affects relative/bare specifier resolution (resolve base). The externalization contract that feeds this hook emits absolute paths (plus `node:` builtins) — both context-independent, and absolute paths now go through `pathToFileURL` → native `import()` regardless of which entry's context the closure captured. `assetFiles` is the same shared object for all entries of a file run.
3. Cross-*file* leakage (the case where contexts genuinely differ) is already handled: `preparePool` unconditionally drops the hook before each file (773bf1f0).

If a real case ever surfaces where a federated chunk passes a context-sensitive specifier through this hook, the right fix would be a current-context indirection in the worker loaders rather than runtime-module logic — happy to revisit then.

website/docs/en/config/test/federation.mdx
```diff
@@ -0,0 +1,66 @@
...
+import { ApiMeta } from '@components/ApiMeta';
+
+<ApiMeta addedVersion="0.10.5" />
+
+- **Type:** `boolean`
+- **Default:** `false`
+- **CLI:** `--federation`
+
+Whether to enable [Module Federation](https://module-federation.io/) compatibility mode.
+
```
9aoy, line 18: Consider adding a short “support scope” note. The current docs don’t make it clear whether Module Federation compatibility applies to browser-like Node environments such as `jsdom` / `happy-dom`, or whether it also covers Rstest Browser Mode. 
For example, it would be helpful to clarify that this option applies to Rstest’s Node-based runner, including `testEnvironment: 'node'` and DOM-simulated environments like `jsdom` / `happy-dom`, while it does not enable or change Rstest Browser Mode (`browser.enabled: true`).

ScriptedAlchemy: Addressed in faaa1464 by adding a support-scope note to the EN/ZH federation docs: this applies to the Node-based runner, including node/jsdom/happy-dom environments, and does not enable or change Browser Mode.

examples/federation/.gitignore
chenjiahan, line 1: Can we remove this file

.github/workflows/test.yml
```diff
@@ -136,6 +136,10 @@ jobs:
       - name: Run Examples Test
         run: pnpm run test:examples
 
+      - name: Run Federation Example Test
+        if: runner.os == 'macOS'
```
9aoy: Can CI run on Windows?

examples/federation/README_zh-cn.md
```diff
@@ -0,0 +1,31 @@
+# Module Federation 示例
```
9aoy: The example does not need a Chinese README.

examples/federation/.gitignore
ScriptedAlchemy: Removed in 309d3b9b. The root .gitignore already covers the relevant local artifacts for this example.

examples/federation/component-app/package.json
chenjiahan, line 1: Should we use Rsbuild instead of Rspack in the examples? This should make the example simpiler

examples/federation/component-app/test/Button.ssr.test.ts
```diff
@@ -0,0 +1,10 @@
+import { expect, test } from '@rstest/core';
+import React from 'react';
+import { renderToString } from 'react-dom/server';
+import Button from '../src/Button.jsx';
+
+test('SSR: Button renders primary', () => {
```
9aoy, line 6: Perhaps we can add a CSR test example.

examples/federation/component-app/package.json
ScriptedAlchemy: Addressed in 7d3e861b. The federation examples now use Rsbuild build configs and @module-federation/rsbuild-plugin for standalone app/remote prebuilds, while the Rstest test configs keep @module-federation/rstest + federation: true for test execution. The raw rspack build CLI configs were removed.

.github/workflows/test.yml
ScriptedAlchemy: Addressed in a8417c42 by running the federation example test on both macOS and Windows in the existing ut matrix.

examples/federation/README_zh-cn.md
ScriptedAlchemy: Addressed in a8417c42 by removing README_zh-cn.md and updating the English README so it no longer links to the removed file.

packages/core/src/runtime/worker/globalSetupWorker.ts
```diff
@@ -73,6 +74,10 @@ const runGlobalSetup = async (data: {
     // Start tracking environment changes
     trackEnvChanges();
 
+    // `mockRuntimeCode.js` gates its Module Federation shims on this
+    // worker-wide flag, so it must be set before any setup code is evaluated.
+    (globalThis as any).__rstest_federation__ = data.federation === true;
```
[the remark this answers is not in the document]
ScriptedAlchemy: Valid, fixed in a8417c42. The global setup worker now clears the federation dynamic-import hook before each setup entry so the shim installs the entry-local resolver instead of reusing the first setup bundle closure.

examples/federation/component-app/test/Button.ssr.test.ts
ScriptedAlchemy: Addressed in a8417c42 by adding a CSR Button example test under a jsdom inline project, while keeping the existing Node SSR/federation tests in their own project.

packages/core/src/runtime/worker/runInPool.ts
```diff
@@ -129,6 +130,18 @@ const preparePool = async (
...
+  // With `isolate: false` a previous file in this worker may have installed
+  // the global dynamic-import fallback (`mockRuntimeCode.js`). Always drop it:
+  // it keeps federation strictly opt-in for non-federation files, and for
+  // federation files the runtime module reinstalls a fresh hook whose
+  // `import()` is bound to the current bundle's vm dynamic-import context
+  // rather than the previous file's.
+  delete (globalThis as any)[RSTEST_DYNAMIC_IMPORT_HOOK];
```
9aoy: We could add an e2e case for **Module Federation + `isolate: false`** to ensure Rstest doesn’t regress this behavior in the future.

ScriptedAlchemy: Added coverage in b917dca8: the federation e2e suite now runs the basic fixture with `--isolate false`, and that fixture has a second test file asserting the worker-reuse path reinstalls the federation dynamic import fallback. Verified locally with `cd e2e && pnpm test federation`.

ScriptedAlchemy:
# RFC: Module Federation Compatibility Mode

This PR adds an opt-in Module Federation compatibility mode for Rstest's
Node-based runner.

Related work:

- Rstest PR: https://github.com/web-infra-dev/rstest/pull/1407
- Module Federation companion PR: https://github.com/module-federation/core/pull/4320

## Summary

Federated tests have two runtimes sharing responsibility for module loading:

- Rstest owns the test worker runtime: mocks, module cache resets, setup files,
  global setup, dynamic import interception, source maps, and worker isolation.
- Module Federation owns remote containers, shared scope initialization, remote
  entry loading, runtime plugins, and async chunk loading.

Without a compatibility layer, Module Federation's Node runtime can evaluate
remote entries and chunks through `fs`, `vm`, `eval`, HTTP fetch, or native
`import()` paths that bypass the Rstest bundle runtime. That breaks the
assumption that test code, mocks, and dynamically loaded chunks all execute
inside the same controlled worker runtime.

The contract in this PR is intentionally small:

- Rstest exposes `federation: true` and `--federation` to prepare the test
  runtime.
- `@module-federation/rstest` supplies the Module Federation build/runtime
  defaults for Rstest builds.
- App builds continue to use the normal app-side Module Federation plugins, such
  as `@module-federation/rsbuild-plugin`.

## Why this needs a first-class mode

Rstest rewrites and bundles test modules so APIs such as `rs.mock`,
`rs.unmock`, `rs.resetModules`, dynamic import interception, and global setup can
work predictably.

In ordinary test bundles, dynamic imports are routed through Rstest's injected
dynamic import hook. That hook preserves the import origin and keeps resolution
aligned with Rstest's mock and module-cache state.

Module Federation changes the graph shape:

- A host can load an HTTP remote entry.
- A host can load a local Node remote entry.
- A remote entry can load more chunks through Rspack's `async-node` runtime.
- `@module-federation/node` runtime plugins can patch chunk handlers.
- Chunks evaluated through `vm` or `eval` do not receive Rstest's usual
  function-argument-injected dynamic import hook.

So a single test can cross from "Rstest-controlled code" into
"Federation-controlled code" and back. Compatibility mode makes that boundary
explicit and stable.

## Runtime model

```mermaid
flowchart LR
  A["Rstest config: federation: true"] --> B["Worker/global setup sets globalThis.__rstest_federation__"]
  B --> C["mockRuntimeCode installs federation shims"]
  C --> D["Rstest mock/runtime ownership remains active"]
  E["@module-federation/rstest"] --> F["ModuleFederationPlugin plus Rstest-safe defaults"]
  F --> G["Remote entries and chunks load in Node-compatible form"]
  D --> H["Federated test executes inside the Rstest worker"]
  G --> H
```

`federation: true` does not configure remotes, exposes, or shared dependencies.
It only enables Rstest's runtime compatibility behavior. The Module Federation
graph stays in the bundler plugin.

## Runtime requirements

### 1. Opt-in behavior

`federation` defaults to `false`. Users opt in through config:

```ts

export default defineConfig({
  federation: true,
});
```

or through the CLI for one-off runs:

```bash
rstest run --federation
```

This keeps existing Node, jsdom, happy-dom, and browser-mode behavior unchanged.
Browser Mode is not enabled by this flag. The flag applies to Rstest's
Node-based runner, including DOM-simulated environments that still execute in
Node workers.

### 2. The runtime flag must be set before user code

The worker sets `globalThis.__rstest_federation__` before evaluating:

- global setup bundles
- setup files
- test bundles
- federation runtime chunks reached from those bundles

The runtime module gates its shims on that flag, so setting it late would make
behavior depend on load order.

### 3. Dynamic import fallback must preserve import origins

Federated `async-node` chunks can call `__rstest_dynamic_import__` from code that
was evaluated by `vm` or `eval`. Those chunks do not receive the normal injected
hook, so compatibility mode provides a `globalThis.__rstest_dynamic_import__`
fallback.

The fallback must:

- load Node builtins directly
- normalize absolute filesystem paths through `pathToFileURL`
- resolve relative specifiers against the injected origin file
- resolve bare specifiers through `createRequire(origin)`
- fall back to native `import()` when no origin is available

This keeps federation chunk loading aligned with the same origin-aware dynamic
import model Rstest uses for non-federated code.

### 4. Worker reuse must not leak stale hooks

With `isolate: false`, multiple test files can reuse the same worker. The
worker now clears the global dynamic import fallback before each file so the
current bundle reinstalls a hook bound to the current file's dynamic-import
context.

That prevents a federation file from leaving a stale fallback behind for later
files, and prevents later federation entries from reusing an origin from an
earlier bundle.

### 5. Federation chunk handlers must not bypass Rstest

Module Federation can install placeholder chunk handlers such as `consumes` and
`remotes` before its runtime fully initializes. In test bundles, those
placeholders can throw too early.

Compatibility mode no-ops only the known throwing placeholders and allows the
real federation runtime to replace them later.

It also protects Rstest-owned chunk-loading handlers. If the Module Federation
Node runtime tries to overwrite handlers such as
`__webpack_require__.f.readFileVm` or `__webpack_require__.f.require`, remote
chunks can be evaluated outside the Rstest runtime. The guard preserves the
first initialized handler and warns on later overwrite attempts.

### 6. Dynamic external imports must still respect mocks

Dynamic `import(request)` of an external module can produce a different module
id from the id patched by a hoisted `rs.mock(request)`. The runtime keeps a
request-to-mocked-id alias so dynamic external imports resolve to the mocked
factory when a mock exists, and to the original module when it does not.

This matters for federated code because remote containers and runtime plugins
often cross module boundaries through dynamic imports.

## Build-side split

The companion `@module-federation/rstest` package is the build-side half for
Rstest test builds. It provides a Rsbuild plugin that:

- injects `ModuleFederationPlugin`
- expects to run under Rstest caller names (`rstest` or `rstest-browser`)
- defaults Rstest builds to Node-compatible federation output unless browser
  target is requested
- disables test-unfriendly defaults such as DTS and manifest generation
- enables async startup
- for Node target, applies `async-node` output defaults, the node runtime plugin,
  CommonJS-compatible container output, and federation externals bypasses

This keeps Rstest from becoming a Module Federation configuration surface. Rstest
only supplies the test-runtime switch. Module Federation remains responsible for
remotes, exposes, shared scope, runtime plugins, and target-specific bundler
defaults.

For standalone application or remote prebuilds, examples should use the normal
app-side federation plugin, such as `@module-federation/rsbuild-plugin`. The
`@module-federation/rstest` plugin is for the Rstest host/test build, not for
replacing every app build configuration.

## Example shape

```ts
import { federation } from '@module-federation/rstest';

export default defineConfig({
  federation: true,
  plugins: [
    federation({
      name: 'main_app',
      remotes: {
        'component-app': 'component_app@http://localhost:3001/remoteEntry.js',
      },
      shared: {
        react: { singleton: true },
      },
    }),
  ],
});
```

## Non-goals

- This does not auto-discover or configure remotes.
- This does not enable Rstest Browser Mode.
- This does not replace `@module-federation/rsbuild-plugin` for app builds.
- This does not make every possible remote-loading shape supported. It defines
  the Rstest and Module Federation runtime contract needed by the supported
  Node-oriented flows.

## Validation in this PR

The PR includes coverage for the compatibility contract:

- config and CLI opt-in
- disabled-by-default behavior
- global setup federation flag propagation
- global dynamic import fallback installation
- absolute path normalization for native `import()`
- origin-aware relative and bare package resolution
- `isolate: false` worker reuse
- `mockRuntimeCode.js` federation shims
- examples for HTTP remotes, local Node remotes, SSR-style tests, jsdom host
  tests, and CSR app builds
- English and Chinese docs for the `federation` option

## Reviewer checklist

- Is the public API small enough: one boolean config option plus one CLI flag?
- Is the split between Rstest runtime compatibility and Module Federation build
  configuration clear?
- Are the runtime shims sufficiently gated behind `federation: true`?
- Do the examples show the intended user path with `@module-federation/rstest`
  for test builds and app-side federation plugins for app builds?
- Are Browser Mode and unsupported remote-loading shapes scoped clearly enough?

9aoy:
commonjs e2e test failed ⬇️

❌ FAIL federation/index.test.ts > federation > should install federation runtime shims when enabled in config (retry x2)
Error: Test failed with exit code 1. Logs:

Rstest v0.10.6
✗ reuse.test.ts (1)
✗ should reinstall the federation dynamic import fallback for reused workers (3ms)
Cannot find module '/Users/runner/work/rstest/rstest/e2e/federation/fixtures/basic/dist/.rstest-temp/reuse-target.mjs' imported from /Users/runner/work/rstest/rstest/packages/core/dist/0~resolveDynamicImport.js

Repro command: 
```
cd e2e && RSTEST_OUTPUT_MODULE=false pnpm exec rstest 'federation/index.test.ts' --testNamePattern 'federation > should install federation runtime shims when enabled in config'
```

https://github.com/web-infra-dev/rstest/actions/runs/28143939302/job/83347403961

ScriptedAlchemy:
@9aoy fixed, thanks for the repro.

Root cause: in the CommonJS / `RSTEST_OUTPUT_MODULE=false` path, a federation chunk could call the global `__rstest_dynamic_import__` fallback without the injected third-argument origin. The fallback then resolved relative imports from the runtime/test bundle context, which is why `./reuse-target.mjs` was looked up under `dist/.rstest-temp` / `packages/core/dist/0~resolveDynamicImport.js` instead of the current test file origin.

The fix adds a worker-global current-origin hook (`__rstest_dynamic_import_origin__`) and refreshes it before global setup, setup files, and test entries. The fallback now uses `origin || globalThis.__rstest_dynamic_import_origin__` for relative/bare specifier resolution, so CJS federation chunks resolve from the active entry.

Verified with the exact repro:

```bash
cd e2e && RSTEST_OUTPUT_MODULE=false pnpm exec rstest 'federation/index.test.ts' --testNamePattern 'federation > should install federation runtime shims when enabled in config'
```

Also rechecked the latest PR head (`edc4ee1`): macOS/Windows unit and e2e checks are green.

9aoy:
@ScriptedAlchemy  hey, the `main` branch now supports modifying Rstest config from an Rsbuild plugin. It hasn’t been released to npm yet, but if you’re interested, you can merge the main branch, give it a try, and help us test this feature 👀  . 

https://rstest.rs/config/build/plugins#modify-rstest-config

packages/core/src/runtime/worker/runInPool.ts
```diff
@@ -358,26 +374,33 @@ const loadFiles = async ({
...
+      testPath: setupTestPath,
       rstestContext,
       assetFiles,
       interopDefault,
+      virtualFsAssetFiles,
     });
   }
 
   tracker?.transition('collect');
+  setFederationDynamicImportOrigin(federation, testPath);
```
9aoy, line 403: **[P1] Preserve the federation import hook across shared runtime entries**

`setFederationDynamicImportOrigin()` deletes `globalThis.__rstest_dynamic_import__` after the setup entries have run. However, setup files and the test entry share a cached `rstest-runtime` chunk. When the test entry requires that chunk, it is not evaluated again, so the deleted hook is never reinstalled.

I reproduced this on the current PR head: after loading a shared runtime from a setup entry and then a test entry, `__rstest_dynamic_import__` remains `undefined`.

Please keep a stable worker-owned hook whose current origin can be updated, or reinstall the hook without relying on the cached runtime chunk to execute again.

packages/core/src/core/plugins/mockRuntimeCode.js
```diff
@@ -1,5 +1,64 @@
...
+              return import(
+                new URL(specifier, url.pathToFileURL(fallbackOrigin)).href,
+                importAttributes
+              );
+            }
+            return import(
+              url.pathToFileURL(
+                module
+                  .createRequire(url.pathToFileURL(fallbackOrigin).href)
+                  .resolve(specifier),
```
9aoy: **[P1] Resolve bare dynamic imports using ESM conditions**

`createRequire(...).resolve(specifier)` applies the `require` conditional-exports conditions, but this fallback implements dynamic `import()` and should use the `import` conditions.

For example, a valid ESM-only package with the following exports currently fails with `No "exports" main defined`:

```json
{
  "exports": {
    "import": "./index.mjs"
  }
}
```

I reproduced this on the current PR head. Please use the same `import.meta.resolve(specifier, parentURL)` semantics as the worker dynamic-import resolver instead of `createRequire().resolve()`.

packages/core/src/pool/index.ts
```diff
@@ -46,8 +46,14 @@ const filterAssetsByEntry = async (
   getAssetFiles: (names: string[]) => Promise<Record<string, string>>,
   getSourceMaps: (names: string[]) => Promise<Record<string, string>>,
   setupAssets: string[],
+  allAssetNames: string[] | undefined,
+  federation: boolean,
 ) => {
-  const assetNames = Array.from(new Set([...entryInfo.files!, ...setupAssets]));
+  const entryAssetNames =
+    federation && allAssetNames
+      ? allAssetNames.filter((name) => !name.endsWith('.map'))
```
9aoy, line 54: **[P2] Avoid copying every compilation asset into every test task**

With federation enabled, each test entry reads all compilation assets instead of only its own dependency graph. `runTests` also prepares every entry's `buildTask` concurrently, so a suite with N test entries can hold N copies of the complete asset map and serialize them independently to workers.

For large federation projects, this makes host memory usage and IPC traffic approach O(entry count × total assets), potentially O(N²).

Could the asset map be shared per build or worker, or could missing federation chunks be requested lazily?

packages/core/src/runtime/worker/loadModule.ts
```diff
@@ -17,6 +17,77 @@ import {
...
+          maybeCallback?: unknown,
+        ) => {
+          const callback =
+            typeof optionsOrCallback === 'function'
+              ? optionsOrCallback
+              : maybeCallback;
+          const content = getAssetContent(assetFiles, filePath);
+
+          if (content !== undefined && typeof callback === 'function') {
+            queueMicrotask(() => callback(null, content));
```
9aoy: **[P2] Preserve the Node fs API contract for virtual assets**

The virtual `readFile` and `readFileSync` implementations always return a string. Native Node returns a `Buffer` when no encoding is provided and only returns a string when an encoding is requested. Code that consumes chunks as Buffers will therefore behave differently in federation mode.

The proxy also leaves `fs.promises.readFile` untouched, so a virtual asset is available through the callback and synchronous APIs but produces `ENOENT` through the Promise API.

Please preserve the native overload semantics, including Buffer results, encoding handling, and `fs.promises.readFile`.

packages/core/src/core/executorCapabilities.ts
```diff
@@ -41,6 +41,7 @@ export const executorCapabilities: Record<
   maxConcurrency: { node: 'supported', browser: 'supported' },
   printConsoleTrace: { node: 'supported', browser: 'supported' },
   disableConsoleIntercept: { node: 'supported', browser: 'supported' },
+  federation: { node: 'supported', browser: 'supported' },
```
9aoy: **[P2] Do not mark federation as supported by the browser executor**

The public documentation defines `federation` as a Node-runner compatibility option and explicitly states that it does not affect Browser Mode. The browser runtime also does not consume this field, but the capability table marks it as `browser: 'supported'` and sends it over the browser wire.

This bypasses the capability validation intended to detect silent no-op runtime fields. Please mark it as `stripped` or `ignored-warn` and remove it from the browser runtime projection, unless the browser executor is updated to consume it.

9aoy (commented):
I found five issues in the current federation implementation. Details are included inline.

packages/core/src/core/plugins/mockRuntimeCode.js
ScriptedAlchemy: Fixed in 78f4f4dc. Bare specifiers now resolve through import.meta.resolve(specifier, parentURL), preserving ESM import export conditions. Added an import-only conditional-exports regression test.

packages/core/src/core/executorCapabilities.ts
ScriptedAlchemy: Fixed in 78f4f4dc. federation is now browser: stripped, removed from the static browser projection, and omitted from BrowserRuntimeConfig. Projection/capability tests cover the contract.

packages/core/src/runtime/worker/runInPool.ts
ScriptedAlchemy: Fixed in 78f4f4dc. setFederationDynamicImportOrigin() now keeps the worker-owned __rstest_dynamic_import__ hook while federation remains enabled and only updates the current origin. It removes both globals when federation is disabled. Added lifecycle regression coverage.

packages/core/src/runtime/worker/loadModule.ts
ScriptedAlchemy: Fixed in 78f4f4dc. The virtual fs proxy now returns Buffer without an encoding, strings with an encoding, and implements the same behavior for callback readFile, readFileSync, and fs.promises.readFile. Added regression coverage for all three APIs.

packages/core/src/pool/index.ts
ScriptedAlchemy: Fixed in 78f4f4dc. Federation tasks no longer eagerly attach the complete compilation asset map during concurrent task preparation. They fetch it through getAssetsByEntry when a worker starts the task, bounding simultaneous full maps by active workers instead of total entry count.

website/docs/zh/config/test/federation.mdx
```diff
@@ -0,0 +1,69 @@
+---
+description: 为 Node 测试启用 Module Federation 兼容模式。
+---
+
+# federation
+
+import { ApiMeta } from '@components/ApiMeta';
+
+<ApiMeta addedVersion="0.10.7" />
```
9aoy: 0.12.0

9aoy:
@ScriptedAlchemy The feature looks good to me within its current scope. One follow-up question: are there plans to support federation compatibility with ESM output (`output.module: true`) and in Rstest Browser Mode? Would supporting these scenarios require any additional functionality from Rstest?

@fi3ework , do you have any additional suggestions or concerns about this PR?

Also, it looks like CI is currently failing. Could you please take a look?

fi3ework:
@ScriptedAlchemy hi Zack, i recently did some refactoring on browser mode to make it more isomorphic with node mode. the last expected refactor has just been merged, so could you please resolve the conflicts again, sorry for another around of rebase.

i'm thinking that in order to better release this feature (which is a huge one) and minimize the need for frequent rebasing on your end, should we mark federation in the config as experimental jsdoc and comment out the page routes in the docs? this way, we can verify it in more repositories before making the API and documentation public to users. 

overall, this would allow us to strike a better balance between development speed and quality. WDYT.

website/docs/zh/config/test/federation.mdx
ScriptedAlchemy: Updated the English and Chinese federation docs to addedVersion 0.12.0 in 13472275. Thanks.

9aoy (approved):

9aoy:
🎉

PR #1412, opened 2026-06-12
fix(core): avoid thenable interop namespaces

9aoy:
## Summary

Fixes dynamic import interop for modules whose default export is a Promise. The interop proxy previously fell through `ns.then` to `default.then`, which made the namespace itself look thenable and caused `Promise.prototype.then called on incompatible receiver [object Module]` when importing native TypeScript ESM modules.

## Checklist


9aoy:
@codex review

9aoy:
@codex review

PR #1414, opened 2026-06-12
fix(core): report untransformed mock API calls

9aoy:
## Summary

This PR fixes a silent failure mode for module mock APIs when calls are not transformed by the built-in Rstest plugin, such as when `rstest` is imported through an alias. Instead of returning placeholder values, plugin-managed APIs now throw a runtime error that explains the required `rstest.*` / `rs.*` call shape and notes that import aliases are not supported for module mock APIs.

## Checklist


9aoy:
@codex review

PR #1418, opened 2026-06-15
fix(core): mock async externals in commonjs output

9aoy:
## Summary

Fixes `{ mock: true }` for CommonJS output when an external ESM module is represented as a promise-returning webpack module. The runtime now waits for async externals to resolve before applying `mockObject`, so dynamic imports like `await import('@rsbuild/core')` receive the mocked module namespace instead of a mocked Promise wrapper.

## Related Links

https://github.com/web-infra-dev/rstest/pull/1415

## Checklist


9aoy:
@codex review

PR #1441, opened 2026-06-20
feat(watcher): enable rspack native watcher

stormslowly:
## Why

Rstest uses the JS file watcher in watch mode because rspack's **native (Rust) watcher** had a macOS flake — stale FSEvents replayed an empty rerun (`expected '' to match 'other.test.ts'`), failing ~7.5% on macos-14 runners.

That root cause is fixed upstream and shipped in **rspack 2.1** (`safeTime` / `FS_ACCURACY` granularity-safe scan + ignored-subtree event filtering). `main` already runs on released `@rsbuild/core ~2.1.4` / `@rspack/core ~2.1.2`, so the fix is already in place — this PR only flips the watcher on, no version bump.

## What

- Enable `experiments.nativeWatcher` in watch mode.
- Raise the default watch `aggregateTimeout` `5 → 100ms` — on macOS the native watcher needs the FSEvent → vnode cache invalidation window to elapse before rspack stats the changed file, otherwise the rebuild reads stale content. Drop the now-redundant per-fixture `aggregateTimeout: 10` overrides in the watch e2e configs so this default applies.

Config-file restart watching is chokidar-side and unrelated to the native watcher; it moved to #1727.

## Validation

Stress-tested on macos-14, where the flake used to surface:

- **Real e2e volume — 50 rounds of the full `pnpm test` suite (~237 files, incl. browser-mode): `0/50` watcher flake.** ([#1520](https://github.com/web-infra-dev/rstest/pull/1520))
- 700+ watch-suite runs across the upstream fix series (rspack 2.1 pre-release): zero watcher flake; released 2.1.x is a superset of those builds.
- Local macOS re-run on 2.1.4: `10/10` green.

The 50-round real-volume run surfaced a separate, **pre-existing** flake in `e2e/no-isolate/fixtures/sharing` (weakly-held mock registry, GC-timing sensitive under load) — unrelated to this PR: the change is watch-mode only and inert under `rstest run`, and that directory is untouched here.


stormslowly:
**Held as draft — blocked on an upstream rspack 2.1 source-map regression.**

CI fails on 3 non-watcher e2e tests (`list/index`, `list/json`, `test-api/timeout`) that assert `includeTaskLocation` source columns. rspack 2.1.0-beta.0 changed the source-map mapping for ESM interop member-call expressions:

```
(0, _core.describe)('test a', ...)   // bundled
```

The call-site column now resolves to the source `(` instead of the `describe`/`it` identifier:

| location | rspack 2.0.x | rspack 2.1.0-beta.0 |
| --- | --- | --- |
| `describe('test a'` (L3) | `3:1` | `3:9` |
| `it('test a-1'` (L4) | `4:3` | `4:5` |
| `it('test a-2'` (L9) | `9:1` | `9:3` |

Root cause confirmed via minimal repro + source-map decode. Not configurable (tested `experiments.sourceImport`, `optimization.concatenateModules`, `devtool`, `output.column`), not yet fixed on rspack main, and rspack 2.1 is required for the native-watcher fix (so downgrading is not an option).

Holding this PR until the upstream regression is fixed; the task locations will then snap back without any test changes here.

PR #1459, opened 2026-06-23
refactor(core)!: make mock factories sync typed

9aoy:
## Summary

This PR makes the public `rs.mock` / `rs.doMock` factory type synchronous-only so TypeScript no longer encourages async mock factories that can run into hoisting and module initialization issues. Partial mock guidance now points users to synchronous `importActual` import attributes, and the e2e coverage was updated to exercise that supported pattern.

## Related Links

https://github.com/web-infra-dev/rstest/issues/1294
https://rstest.rs/guide/migration/vitest#mock-async-modules

## Checklist


fi3ework (commented):
**Also drop the async tolerance in the runtime — throw instead of silently half-supporting it.**

The type now forbids async factories, but `mockRuntimeCode.js` still accepts them in the function-factory arm:

```js
const res = modFactory();
if (isPromise(res)) {
  __webpack_module__.exports = res;   // async factory still slips through (JS / @ts-expect-error / as any)
  return;
}
```

Result is the exact footgun this PR targets: works via `await import()`, but silently breaks via static `import { x }` with `x is not a function`. Since this is already breaking, fail loud:

```suggestion
        const res = modFactory();

        if (isPromise(res)) {
          throw new Error(
            `[Rstest] An async mock factory is not supported. ` +
              `Use a sync factory; to keep part of the original module, ` +
              `import it with \`with { rstest: 'importActual' }\` and spread it in.`,
          );
        }
```

Don't touch the other `isPromise` use (~L213–227, `{ spy/mock: true }` async **original** module) or the L47 helper. Bonus: a `// @ts-expect-error` test so the type can't regress to `MaybePromise`.

PR #1460, opened 2026-06-23
fix(core): avoid excluding test directories from coverage

### LINKED ISSUE #1458: [Bug]: Istanbul coverage of untested files crashes ("cannot find magic value") for any package under a `test/` directory
### Version

```sh
System:
 - OS: macOS 26.5.1
 - CPU: (10) arm64 Apple M1 Pro
 - Memory: 202.44 MB / 32.00 GB
 - Shell: 5.9 - /opt/homebrew/bin/zsh

Browsers:
 - Chrome: 149.0.7827.156
 - Edge: 149.0.4022.80
 - Firefox: 152.0.1
 - Safari: 26.5

Tested on: `@rstest/core` + `@rstest/coverage-istanbul` v0.10.6, Node 22
```

### Details

`rstest run --coverage` (istanbul) exits **1** with `Error: cannot find magic value`
for an untested source file whenever the package's absolute path contains a
`test` segment (e.g. a monorepo lib named `test`, `packages/test/`). 

A byte-for-byte identical package under `lib/` name instead reports the file at 0% and exits 0.

This is a unique case, but we have a module in the monorepo named `test` which contains non-spec files as well. Seems like renaming would work but would be nice if there's an alternative easier way to properly enable coverage for that package without renaming (to avoid churn)

### Reproduce link

https://github.com/giavinh79/rstest-coverage-test-dir-collision

### Reproduce Steps

**See reproduction repro above.**

**Steps:**
1. pnpm install
2. pnpm run repro:fail   # packages/test  -> exits 1, "cannot find magic value"
3. pnpm run repro:pass   # packages/lib   -> exits 0  (identical files, different dir name)

**Expected:**
Behaves like the `lib/` package (`repro:pass`): report the untested file at 0% and exit 0.
The two packages differ only in their parent directory name.

9aoy:
## Summary

Fixes Istanbul coverage for projects whose absolute path contains a `test` segment, such as `packages/test`. Rstest no longer excludes `test` directories from coverage by default, which avoids dropping untested source files before Istanbul instrumentation metadata is available. The default still excludes `node_modules`, `__tests__`, `__mocks__`, declaration files, and `*.test` / `*.spec` files; users who want directory-based filtering can add `test/**` or `**/test/**` explicitly.

## Related Links

close: https://github.com/web-infra-dev/rstest/issues/1458

## Checklist


9aoy:
@codex review

PR #1469, opened 2026-06-24
feat(core): add teamcity reporter

omarshibli:
## Summary

Adds a built-in `teamcity` reporter that emits [TeamCity service messages](https://www.jetbrains.com/help/teamcity/service-messages.html) to stdout, so TeamCity (and compatible CI servers) populate the **Tests** tab directly from the build log — the service-message counterpart to the existing `junit` (XML file) and `github-actions` (annotations) reporters.

Opt-in via config:

```ts
export default defineConfig({
  reporters: ['default', 'teamcity'],
});
```

## Behavior

- Emits in a single `onTestRunEnd` pass over the result tree (same shape as `junit`/`github-actions`), so `started`/`finished` messages are always balanced without tracking state.
- Each test file → a `testSuite` (relative path as `name`/`flowId`); cases → `testStarted`/`testFinished`, `testIgnored` (skip/todo), or a single `testFailed`.
- Comparison failures carry `expected`/`actual` via `type='comparisonFailure'`.
- Multiple errors per test are aggregated into one `testFailed` (TeamCity records only the first `testFailed` per test).
- A file that fails to load with no cases is reported as a synthetic `file setup` failed test.
- Values escaped per the TeamCity spec; ANSI stripped; stacks sourcemapped via the shared `parseErrorStacktrace`/`formatStack` helpers.

Example output:

```text
##teamcity[testSuiteStarted flowId='test1.test.ts' name='test1.test.ts']
##teamcity[testStarted flowId='test1.test.ts' name='should pass']
##teamcity[testFinished duration='1' flowId='test1.test.ts' name='should pass']
##teamcity[testStarted flowId='test1.test.ts' name='should fail']
##teamcity[testFailed actual='hi' details='...' expected='hii' flowId='test1.test.ts' message='...' name='should fail' type='comparisonFailure']
##teamcity[testFinished duration='2' flowId='test1.test.ts' name='should fail']
##teamcity[testIgnored flowId='test1.test.ts' name='should skip']
##teamcity[testSuiteFinished flowId='test1.test.ts' name='test1.test.ts']
```

## Tests & docs

- Unit: `packages/core/tests/reporter/teamcity.test.ts`
- E2E: `e2e/reporter/teamcity.test.ts` (+ fixture), running the real CLI with `--reporter teamcity`
- Docs: `website/docs/{en,zh}/guide/basic/reporters.mdx`

## Open questions for reviewers

- **`details` content:** this reporter formats the full sourcemapped stack into `details` (TeamCity's details pane is built for full traces). `junit`/`github-actions` use only the top frame — happy to switch to top-frame for parity if preferred.

### DIFF
--- a/e2e/reporter/fixtures/teamcity.test.ts
+++ b/e2e/reporter/fixtures/teamcity.test.ts
@@ -1,0 +1,15 @@
+import { describe, expect, it } from '@rstest/core';
+
+describe('Teamcity test', () => {
+  it('should pass', () => {
+    expect(1 + 1).toBe(2);
+  });
+
+  it('should fail', () => {
+    expect('hi').toBe('hii');
+  });
+
+  it.skip('should skip', () => {
+    expect(1 + 1).toBe(3);
+  });
+});
--- a/e2e/reporter/teamcity.test.ts
+++ b/e2e/reporter/teamcity.test.ts
@@ -1,0 +1,40 @@
+import { expect, it } from '@rstest/core';
+import { runRstestCli } from '../scripts';
+
+it('teamcity', async () => {
+  const { cli, expectLog } = await runRstestCli({
+    command: 'rstest',
+    args: ['run', 'teamcity', '--reporter', 'teamcity'],
+    options: {
+      nodeOptions: {
+        cwd: __dirname,
+      },
+    },
+  });
+
+  await cli.exec;
+  expect(cli.exec.process?.exitCode).toBe(1);
+
+  const logs = cli.stdout.split('\n').filter(Boolean);
+
+  expectLog(
+    "##teamcity[testSuiteStarted flowId='fixtures/teamcity.test.ts' name='fixtures/teamcity.test.ts']",
+    logs,
+  );
+  expectLog(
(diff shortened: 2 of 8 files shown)


fi3ework:
thanks, this is my first time learning about TeamCity. given that JUnit already provides basic support for TeamCity, and TeamCity hasn't reached the same level of adoption as GitHub Actions, i'm concerned that this could lead to an infinite expansion of reporters in the future (teamcity2, teamcity3, etc.).

so, if possible, i would prefer to make this TeamCity reporter independently maintained, and we can provide a link in the official documentation. what do you think?

omarshibli:
> thanks, this is my first time learning about TeamCity. given that JUnit already provides basic support for TeamCity, and TeamCity hasn't reached the same level of adoption as GitHub Actions, i'm concerned that this could lead to an infinite expansion of reporters in the future (teamcity2, teamcity3, etc.).
> 
> 
> 
> so, if possible, i would prefer to make this TeamCity reporter independently maintained, and we can provide a link in the official documentation. what do you think?

Sounds good, is there a specific repo for rstest external contributions?

fi3ework:
@omarshibli no, no yet, you can just create under your handle.

PR #1470, opened 2026-06-24
chore: add api-doc-sync skill to guard API doc signature drift

fi3ework:
## Summary

Adds an `api-doc-sync` agent skill that defends against **API doc signature drift** — the `**Type:**` / `**类型：**` blocks in `website/docs/**/api/**` are hand-written, curated copies of the real exported types, with no generation or compiler check behind them, so they can be wrong the moment they are authored (not only when the type later changes). This came directly out of #1468, where the documented `describe`/`test` signatures had silently dropped an overload.

The skill is split into a deterministic half and a judgement half:

- **Deterministic (`scripts/check-type-blocks.mjs`):** extracts every `**Type:**` block from the en pages, pairs each with its zh counterpart, and asserts the signatures are structurally identical (label and translated `//` comments may differ). Exits non-zero on any mismatch / missing counterpart / block-count difference, so it can be wired into CI or pre-push as a hard gate. Runs clean on `main` today (149 blocks across 12 pages) and catches a structural en/zh divergence.
- **Judgement (`SKILL.md` procedure):** for each changed symbol, find the source-of-truth type in `packages/core/src/types/` (or the built `.d.ts`), turn every documented call form into `tsc` compile assertions (positive forms + `@ts-expect-error` for forms the docs call invalid), and explicitly diff the source's overload set against the documented one to catch *missing* overloads — the exact failure mode from #1468 that a positive-only check misses.

The core rule baked into the skill is **verify, never recall**: ground every signature in source + `tsc`, never in memory or the doc's own prose.

### Trigger (so it isn't memory-dependent)

- `development` skill gains a routing line: changed a public type in `packages/core/src/types/`, or edited a `**Type:**` block → run `api-doc-sync`.
- Registered in the skills table (`AGENTS.md`; `CLAUDE.md` is a symlink to it).

### Deliberately out of scope

- **Generating** signatures from source: would make drift impossible but discards the curated/simplified doc style (friendly names, omitted internal generics). Noted as a non-goal.
- **Wiring the parity script into lefthook/CI as a blocking gate:** the script is ready for it, but adding a new blocking gate is a separate decision — happy to follow up if wanted.
- **Typechecking doc *example* snippets** (twoslash): complementary, larger surface, but would not have caught the #1468 signature bug. Separate effort.

## Related Links

- Follow-up to #1468

## Checklist

- [x] Tests updated (or not required). _Skill tooling; `check-type-blocks.mjs` verified manually against `main` (clean) and a drift fixture (flagged)._
- [x] Documentation updated (or not required). _Skill registered in `AGENTS.md` and routed from the `development` skill._


website/docs/en/api/runtime-api/test-api/hooks.mdx
```diff
@@ -55,6 +55,8 @@ afterAll(async (ctx) => {
 
 - **Type:** `(fn: (ctx: TestContext) => void | Promise<void>, timeout?: number) => void`
 
+The `ctx` argument is a [`TestContext`](/api/runtime-api/test-api/test#testcontext).
```
9aoy: This type description looks a bit repetitive.

fi3ework: only mention once now.

9aoy (approved):

PR #1474, opened 2026-06-25
ci: use GitHub Chrome for browser tests

9aoy:
## Summary

This PR reduces CI setup work for browser tests by using the Chrome already available on GitHub-hosted runners instead of installing Playwright Chromium. Browser e2e helpers and browser examples inject `browser.providerOptions.launch.channel = 'chrome'` when `process.env.CI` is set, while the WebKit fixture still installs WebKit explicitly. It also exposes `--browser.providerOptions.*` so tests can pass provider-specific launch options through CLI without changing Playwright provider/runtime code.

## Checklist


fi3ework (approved):

fi3ework:
👍🏻

PR #1522, opened 2026-07-09
fix(core): preserve test-file dispatch order under isolate: false

stormslowly:
## Summary

Fixes an intermittent `isolate: false` failure: a `globalThis` / shared-module
value written by an earlier test file was sometimes `undefined` in a later file,
but only under full-suite load.

With `pool.maxWorkers: 1`, files run in the order they reach the pool. `runTests`
awaited each file's `buildTask` inline before submitting it, and `buildTask`'s
asset-assembly latency varies (skipped entirely when memory is tight) — so under
memory pressure a later file could be submitted first and run before an earlier
one, breaking cross-file sharing.

Fix: build all tasks first, then submit them to the pool in entry order. Asset
assembly still runs concurrently — only the submit order is pinned.

- **Before:** submit order = whichever `buildTask` resolves first ≠ entry order.
- **After:** submit in entry order → deterministic execution order.

## Related Links

- Fixtures / prior art: #1373, #1376

## Checklist

- [x] Tests updated (or not required). Covered by e2e `e2e/no-isolate/moduleSharing.test.ts`.
- [x] Documentation updated (or not required). Internal scheduling only.

### DIFF
--- a/packages/core/src/pool/index.ts
+++ b/packages/core/src/pool/index.ts
@@ -492,13 +492,9 @@ export const createPool = async ({
       });
       const setupAssets = setupEntries.flatMap((entry) => entry.files || []);
 
-      const results = await Promise.all(
-        entries.map(async (entryInfo, index) => {
-          const traceArgs = {
-            project: projectName,
-            testPath: entryInfo.testPath,
-          };
-          const task = await traceSpan(
+      const tasks = await Promise.all(
+        entries.map((entryInfo, index) =>
+          traceSpan(
             'host:build-task',
             'host',
             () =>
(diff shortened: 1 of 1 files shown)


stormslowly:
Closing — already fixed on `main`.

#1531 (perf-first test-file ordering) landed the same root-cause fix in `pool/index.ts`: its "sequential dispatch gate" serializes the pool enqueue so a later file whose `buildTask` finishes first can't jump ahead of an earlier one — the same reorder this PR addressed, with better pipelining (asset reads stay overlapped; only the enqueue is serialized).

#1523 additionally reworked the no-isolate sharing fixtures to be order-independent, so the specific files this PR referenced no longer exist.

Superseded on both counts.

PR #1536, opened 2026-07-09
fix(core): report running test case as failed on worker crash

### LINKED ISSUE #1535: [Bug]: A test case that crashes its worker is not reported as a failed test
### Version

```sh
System:
  OS: macOS 15.0.1 (arm64)
  CPU: (14) Apple M3 Max
npmPackages:
  @rstest/core: 0.11.0
```

### Details

When a test worker process exits unexpectedly while a test is running — for example a native addon segfaults, an OOM kill, or any abrupt/uncatchable process death — rstest marks the affected **file** as failed, but the **test case that was executing is never reported as a failed test**. It disappears from the test-level counts.

In the minimal repro below, one worker is terminated mid-test with `process.abort()` (a stand-in for any uncatchable worker death, such as a native addon crash). `npx rstest run` prints:

```
 ✓ tests/passing.test.ts (1)
 ✗ tests/worker-crash.test.ts (0)

Summary of all failing tests:
 FAIL  tests/worker-crash.test.ts
Error: Worker exited unexpectedly (code=null, signal=SIGABRT)

Maybe relevant test case: "worker dies mid-test" which is running when the error occurs.

 Test Files 1 failed | 1 passed (2)
      Tests 1 passed
```

Observations:

- The crashed file is reported as `✗ tests/worker-crash.test.ts (0)` — 0 tests.
- The footer shows `Test Files 1 failed | 1 passed`, but `Tests 1 passed` — the `worker dies mid-test` case that was running when the worker exited is not counted as a failed test anywhere in the test totals. Only the passing test is counted.
- rstest already identifies which case was running: it prints `Maybe relevant test case: "worker dies mid-test" which is running when the error occurs.`

So the run correctly exits non-zero, but the failure is attributed only at the file level; the specific test case that caused the worker to exit is left uncounted.

### Expected behavior

The test case that was executing when the worker exited unexpectedly should be reported as a **failed test case** and reflected in the `Tests` totals (e.g. `Tests 1 failed | 1 passed`), attributed to that specific case — while the rest of the run completes and reports the other files normally, and the process exits non-zero.

In other words, an unexpected worker exit should surface as a normal failing test case, not only as a file-level failure with the running case dropped from the counts.

### Reproduce link

Self-contained — no external repo required. The four files below reproduce it.

### Reproduce Steps

1. Create a project with these files:

`package.json`

```json
{
  "name": "rstest-worker-exit-repro",
  "private": true,
  "devDependencies": {
    "@rstest/core": "latest"
  }
}
```

`rstest.config.ts`

```ts

export default defineConfig({
  include: ['tests/**/*.test.ts'],
});
```

`tests/passing.test.ts`

```ts

test('a normal passing test', () => {
  expect(1 + 1).toBe(2);
});
```

`tests/worker-crash.test.ts`

```ts
import { test } from '@rstest/core';

test('worker dies mid-test', () => {
  // Uncatchable abrupt process death, like a native addon segfault.
  process.abort();
});
```

2. Run `npm install`.

3. Run `npx rstest run`.

4. Observe the footer: `Test Files 1 failed | 1 passed`, but `Tests 1 passed` — the `worker dies mid-test` case is not counted as a failed test, and the crashed file is reported with `(0)` tests.

fi3ework:
## Summary

When a worker process exits unexpectedly while a test is running (a native addon crash, an OOM kill, or any abrupt process death), rstest marked only the **file** as failed — the test case that was executing vanished from the `Tests` counts and the crashed file was reported with `(0)` tests, even though rstest already knew which case was running.

This makes an unexpected worker exit surface as a normal failing test case: the running case(s) are synthesized as failed `TestResult`s and included in the file's results, so they appear in the `Tests` totals (`Tests 1 failed | 1 passed`) attributed to that specific case. The crash error is attributed to those case(s) and kept off the file-level result so the failing-tests summary does not print it twice. When no case was running (crash during setup/import), the error stays at the file level as before.

## Related Links

- Closes #1535

## Checklist

### DIFF
--- a/e2e/worker/index.test.ts
+++ b/e2e/worker/index.test.ts
@@ -44,6 +44,24 @@ describe('test worker behavior', () => {
     expect(cli.log).toContain(marker);
   });
 
+  it('should report the running test case as failed when worker exits unexpectedly', async () => {
+    const { expectExecFailed, cli } = await runRstestCli({
+      command: 'rstest',
+      args: ['run', 'worker.panic.test.ts'],
+      options: {
+        nodeOptions: {
+          cwd: fixtureDir,
+        },
+      },
+    });
+
+    await expectExecFailed();
+    // The case that was running at crash time is attributed as a failed test
+    // case rather than silently dropped from the counts (#1535).
+    expect(cli.log).toContain('should crash the worker process');
+    expect(cli.stdout).toMatch(/Tests\s+1 failed/);
+  });
+
   it('should handle unhandledRejection error correctly', async () => {
     const { expectExecFailed, expectStderrLog } = await runRstestCli({
       command: 'rstest',
--- a/packages/core/src/pool/index.ts
+++ b/packages/core/src/pool/index.ts
@@ -226,13 +226,18 @@ const buildTask = async ({
  * Convert a worker crash or pool error into a fail-status `TestFileResult`.
  * Enriches the error with context about which test cases were running at the
  * time of the crash (if any).
+ *
+ * Returns the file result plus the synthetic `crashedResults` (the cases that
+ * were running at crash time). The caller replays those through the live
+ * `onTestCaseResult` reporter hook so incremental reporters stay consistent
+ * with the final totals — they are already included in `fileResult.results`.
  */
 const workerErrorToResult = (
   err: unknown,
   testPath: string,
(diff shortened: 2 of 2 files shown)


fi3ework:
@codex review

PR #1539, opened 2026-07-09
fix: re-export AsymmetricMatchersContaining and Matchers types

EskiMojo14:
## Summary

Re-exporting AsymmetricMatchersContaining and Matchers allows them to be augmented directly:

```ts
declare module "@rstest/core" {
  interface Matchers<T> {
    toBeEven: () => void;
  }
  interface AsymmetricMatchersContaining {
    even: () => any
  }
}

expect(2).toBeEven()
expect({ count: 2 }).toEqual({ count: expect.even() })
expect({ count: 3 }).toEqual({ count: expect.not.even() })
```

This fixes `expect.not.even()` not being exposed (if `Assertion` is augmented instead) and is also useful for libraries that only want to expose matchers as either symmetric or asymmetric, but not both (e.g. [mix-n-matchers](https://eskimojo14.github.io/mix-n-matchers/))

## Checklist

### DIFF
--- a/packages/core/src/index.ts
+++ b/packages/core/src/index.ts
@@ -98,10 +98,12 @@ export type { Rspack } from '@rsbuild/core';
 export type {
   Assertion,
   DescribeAPI as Describe,
+  AsymmetricMatchersContaining,
   ExpectStatic,
   ExtendConfig,
   ExtendConfigFn,
   Fixtures,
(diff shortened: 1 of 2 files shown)


9aoy (approved):

9aoy:
👍

PR #1545, opened 2026-07-10
feat(core): support --onlyFailures to re-run failed test files

fi3ework:
## Summary

Adds `--onlyFailures` (`-f`), a boolean config + CLI flag that re-runs only the test files that failed on the previous run, backed by the persistent results cache introduced in #1531.

- **Selection.** When enabled, Rstest reads each file's last-known pass/fail state from the results cache and narrows execution to the failed files. Selection is per **file** (any failed test re-runs the whole file); newly added / never-run files are not selected.
- **Empty state.** If nothing failed last run — or there is no cache yet — there is nothing to re-run, so Rstest runs the full suite instead of nothing and prints a notice (`pytest`'s default, not the "run nothing" behavior). It also prints a one-line `running X of Y test files (Z deselected)` summary when it narrows.
- **Guards.** Ignored in watch mode (warned, not errored) since watch already re-runs edited files and offers a run-failed shortcut, so a shared config stays usable. An explicit file filter always wins over failure history. The cache write is skipped for `testNamePattern` and `bail`-aborted runs so a partial run can't poison the failed-first record.

Selection reuses the sequencer's existing failure hints (single source of truth); the `run()` flow is split into build-stats → selection → execute so the global empty-state decision can see every project's candidate set before ordering or running anything.

## Related Links

- Builds on #1531 (perf-first ordering + results cache)

## Checklist


packages/core/src/core/runTests.ts
```diff
@@ -717,38 +692,132 @@ export async function runTests(context: Rstest): Promise<void> {
...
+    // after a bail-aborted or `testNamePattern` run that skipped its write —
+    // could wrongly deselect a file that just failed in memory.
+    const isExplicitlyScoped = !!(
+      fileFilters?.length || context.fileFilters?.length
+    );
+    if (
+      context.normalizedConfig.onlyFailures &&
+      !isWatchMode &&
+      mode !== 'on-demand' &&
+      !isExplicitlyScoped
```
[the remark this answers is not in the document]
fi3ework: Fixed in 74e028d6f. `--onlyFailures` selection is now skipped whenever `--changed`/`--related` is active (`context.relatedMode` guard), which covers the forceRerunTriggers path by construction — the forced full-suite rerun is no longer narrowed by failure history. A one-line warning is emitted for the ignored option, consistent with watch mode and explicit file filters, and the docs now describe the precedence rules.

packages/core/src/core/runTests.ts
```diff
@@ -251,6 +252,34 @@ export async function runTests(context: Rstest): Promise<void> {
...
 
+  // `onlyFailures` applies only to a plain, full run: any other scoping
+  // mechanism wins over failure history. Watch has its own run-failed shortcut
+  // and refreshes the failed set as you edit. `--changed`/`--related` scope by
+  // relevance to a change — and their forceRerunTriggers path deliberately
+  // clears all file filters to force a full-suite rerun, which failure
+  // selection must not narrow back down. Explicit file filters are the user
+  // naming exactly what to run. Warn once and ignore (rather than erroring) so
+  // a shared config carrying `onlyFailures` stays usable everywhere.
+  if (context.normalizedConfig.onlyFailures) {
```
9aoy, line 263: Maybe we can simplify this by turning the previous failed files into internal file filters and reusing the existing discovery/build/run pipeline, instead of adding a second filtering phase over `projectPlans`.

One thing to watch out for is that the filter should probably be **project-aware**. Since the cache key is `project + "\0" + relativePath`, flattening failed entries into a single global `string[]` could accidentally select the same relative test path from another project. A safer shape might be something like `Map<projectName, failedTestPaths[]>`, so each project only receives its own failed-file filters.

This might also make the `--shard` behavior easier to reason about, depending on whether we want to apply the failed-file filter before or after sharding.

fi3ework: Thanks — good points. Some context on why selection is a second phase over `projectPlans` rather than internal file filters:

The current implementation is already project-aware: selection runs per project and matches against `sequenceKey(project.name, rootPath, testPath)`, so the same relative path from another project is never accidentally selected.

The phase-2 shape buys three semantics that a filter-first approach doesn't get for free:

1. **Stale entries self-heal.** A previously-failed file that was since deleted/renamed simply drops out when intersecting with the discovered candidate set. As internal file filters, stale paths would match nothing and fall into the "No test files found" error path instead of gracefully running everything.
2. **The global empty state.** "Nothing failed → run everything with a notice" is only decidable once every project's candidate set is known — that's why `run()` was split into build-stats → selection → execute.
3. **Uncovered-project fallback.** Browser project results aren't recorded in this cache, so those projects fall back to running all their files rather than being silently deselected by a global filter.

Also, the existing `fileFilters` plumbing is a global `string[]` applied identically to every project (glob + stats filter), so the `Map<projectName, paths[]>` shape you suggested — which I agree is the right one — would be new per-project machinery through that pipeline rather than pure reuse.

That said, filter-first has one real upside the current shape lacks: narrowing at glob time would skip compiling passed files entirely, which matters for large repos. I'd like to keep that as a follow-up, and use the project-keyed map shape when doing it.

On `--shard`: the current order (shard resolved upstream, failure filter applied within each shard) keeps a file's shard assignment deterministic across runs, so the full shard matrix still re-runs every failed file. Filtering before sharding would redistribute failed files across shards — better balance, but assignments diverge from the original run. Both seem defensible; happy to revisit if you feel strongly.

PR #1585, opened 2026-07-16
fix(core): support object URLs for DOM blobs

9aoy:
## Summary

- Fix `URL.createObjectURL` rejecting jsdom and happy-dom `Blob`/`File` instances when the Node `URL` and DOM blob implementations come from different realms.
- Add a jsdom-compatible `URL` wrapper that converts jsdom blob storage to a Node `Blob`, and expose happy-dom's existing compatible `URL`.
- Add end-to-end coverage for jsdom and happy-dom with environment `Blob`, `File`, and Node `Blob` inputs.

This is an internal environment compatibility fix with no public API or configuration changes.

## Checklist

### DIFF
--- a/e2e/dom/fixtures/test/domScriptUrl.test.ts
+++ b/e2e/dom/fixtures/test/domScriptUrl.test.ts
@@ -1,0 +1,20 @@
+import { expect, test } from '@rstest/core';
+
+test('should expose object URLs to scripts in the jsdom realm', () => {
+  const script = document.createElement('script');
+  script.textContent = `
+    document.documentElement.dataset.objectUrl = URL.createObjectURL(
+      new Blob(['script blob'], { type: 'text/plain' }),
+    );
+    document.documentElement.dataset.urlSearchParamsRealm = String(
+      new URL('https://example.test/?key=value').searchParams
+        instanceof URLSearchParams,
+    );
+  `;
+  document.head.appendChild(script);
+
+  const url = document.documentElement.dataset.objectUrl!;
+  expect(url).toMatch(/^blob:nodedata:/);
+  expect(document.documentElement.dataset.urlSearchParamsRealm).toBe('true');
+  URL.revokeObjectURL(url);
+});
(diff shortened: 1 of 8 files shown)


9aoy:
Addressed all three review suggestions in `920ea1e8`:

- Track object URLs created by both jsdom and happy-dom and revoke any remaining URLs during environment teardown.
- Detect jsdom blobs through their shared private implementation symbol so blobs from iframe/other jsdom realms are converted correctly.
- Install the compatible URL constructor on `dom.window` during `beforeParse`, with an end-to-end test that executes `URL.createObjectURL` inside the jsdom script realm.

Validation: core tests (64 files / 601 tests), DOM E2E in ESM and CJS modes (17/17 each), type/lint checks, unused checks, formatting, and spelling all pass.

9aoy:
Addressed the follow-up review in `7517fea7`:

- The user's `beforeParse` callback now runs first, and Rstest augments the resulting `window.URL` constructor instead of replacing it. A direct environment setup test covers a customized URL constructor because function-valued environment options cannot be cloned through the normal worker E2E path.
- The implementation now adds only the missing `createObjectURL`/`revokeObjectURL` static methods to jsdom's native URL constructor. This preserves jsdom's URL/URLSearchParams realm invariant; the DOM-script E2E asserts it explicitly.

Validation: core tests (65 files / 602 tests), DOM E2E in ESM and CJS modes (17/17 each), type/lint checks, unused checks, formatting, and spelling all pass.

9aoy:
Addressed the latest follow-up in `0392e714`:

- Install jsdom's `URL` and `URLSearchParams` together in the worker global and restore both during teardown.
- Add a normal test-module assertion for `new URL(...).searchParams instanceof URLSearchParams`, in addition to the DOM-script realm assertion.

Validation: core 65 files / 602 tests, ESM and CJS DOM E2E 17/17 each, plus type/lint/unused/format/spell checks.

Issue #1656, opened 2026-07-31
perf(playwright): keep the browser alive across test files in the same worker


9aoy:
## Problem

`@rstest/playwright@0.11.5` closes its cached browser after the browser fixture has been idle for 1 second (`BROWSER_IDLE_CLOSE_DELAY = 1000`).

For E2E suites that start or rebuild a dev server in `beforeAll`, the gap between files is often longer than one second. The idle timer then closes the browser before the next file requests the fixture, so Chrome is relaunched for almost every test file.

This differs from `@playwright/test`, which normally keeps one browser process alive for the lifetime of a worker and creates an isolated context/page per test.

## Reproduction

Rspress is migrating its E2E suite in https://github.com/web-infra-dev/rspress/pull/3569.

With five representative files and one worker:

```bash
DEBUG='pw:browser*' CI=1 rstest run \
  --config rstest.e2e.config.mts \
  --pool.maxWorkers 1 \
  e2e/fixtures/title-number/index.test.ts \
  e2e/fixtures/custom-icon/index.test.ts \
  e2e/fixtures/custom-home-footer/index.test.ts \
  e2e/fixtures/theme-css/index.test.ts \
  e2e/fixtures/heading-title/index.test.ts
```

The debug output contains five separate `pw:browser <launching>` events: one browser launch per file. The equivalent native Playwright run launches one browser per worker.

On the full Rspress CI suite (70 files / 221 tests), the latest Rstest run took 6m24s while the Rstest build phase was only 180ms. Repeated Chrome startup appears to account for much of the remaining runtime difference from native Playwright.

## Suggested behavior

Prefer keeping the cached browser alive until worker shutdown, while continuing to create and close a fresh browser context for each test.

If an unconditional worker-lifetime cache is undesirable, exposing a Playwright option such as one of the following would let integration suites choose the lifecycle explicitly:

```ts
{
  browserIdleTimeout: Infinity,
  // or
  reuseBrowser: true,
}
```

A configurable idle timeout would also be sufficient, provided the browser is still reliably closed during worker teardown and failure/debug paths.

PR #1661, opened 2026-07-31
feat(core): support scoped fixtures

9aoy:
## Summary

Add `test`, `file`, and `worker` fixture scopes to `test.extend`, including the builder form `test.extend(name, options, fixture)` with `onCleanup`.

The scoped API is intentionally narrow:

- the existing object form remains test-scoped and keeps its `auto` support;
- file- and worker-scoped fixtures use the named builder form, must be declared at the test file's top level, and initialize lazily;
- file/worker scopes do not support `auto` and cannot be overridden;
- a fixture may depend only on fixtures with an equal or longer lifetime, and file/worker fixtures cannot access `TestContext`;
- omitting options in the builder form creates a test-scoped fixture.

These constraints keep the scoped dependency graph stable for its owning lifetime and avoid ambiguous object-form tuples. In particular, an object-form value such as `[value, { scope: 'file' }]` remains ordinary test-scoped array data rather than becoming scope options.

Worker scope follows the actual worker lifetime. Under `isolate: false`, it can span files; when a watch rebuild changes `buildId`, the host stops the old worker through the bounded cleanup path before dispatching the new build.

Node and browser integration both:

- clean file fixtures after suite hooks and worker fixtures before global teardown;
- bound file/worker cleanup at the host, so a never-settling cleanup cannot hang the command;
- include cleanup failures in the finalized run verdict;
- preserve worker-cleanup console output;
- include file cleanup time in the file duration;
- release scoped instance/context references at cleanup, including failure paths.

Cleanup-only execution is not part of the coverage contract. A cleanup callback must also return/await work it owns; detached async work is outside the fixture lifecycle.

This PR does not change Playwright browser reuse behavior.

## Example

```ts
import { test } from '@rstest/core';

const apiTest = test
  .extend('port', { scope: 'worker' }, 5000)
  .extend('server', { scope: 'file' }, async ({ port }, { onCleanup }) => {
    const server = await startServer(port);
    onCleanup(() => server.close());
    return server;
  })
  .extend('client', async ({ server }, { onCleanup }) => {
    const client = createClient(server.url);
    onCleanup(() => client.close());
    return client;
  });

apiTest('fetches a profile', async ({ client }) => {
  const profile = await client.get('/profile');
  expect(profile.name).toBe('Alice');
});
```

`client` is test-scoped because its builder omits `scope`; it is recreated and cleaned up for every test attempt.

## Validation

- `pnpm --filter @rstest/core test` — 681 tests passed
- Browser, Browser UI, and Playwright unit suites — 206 tests passed
- Node E2E — scope order, both isolation modes, cleanup verdict, host timeout, isolated environment teardown order, top-level restriction, file duration, watch worker replacement, global teardown order, and published declarations
- Browser and Playwright E2E — 19 passed; 6 headed-only cases skipped by the local environment
- `pnpm lint`, `pnpm typecheck`, and `pnpm check-unused`
- API documentation parity plus compile probes against emitted declarations
- CPU benchmark averages: compile 64.24ms, runner 78.37ms, integration 78.49ms locally
- Memory benchmark workload: 60/60 files passed; Linux memory metrics remain CI-owned

## Related links

- Enables the follow-up browser-lifecycle work in https://github.com/web-infra-dev/rstest/issues/1656

## Checklist

- [x] Tests updated
- [x] Documentation updated


9aoy:
Closing this PR in favor of smaller, independently reviewable changes.

Follow-up plan:

1. Add the named test-scoped builder form with `onCleanup` end to end:
   - `test.extend(name, fixture)`
   - keep the existing object form and its semantics unchanged
   - cover core, browser mode, and the Playwright wrapper
2. Add file-scoped fixtures as a complete lifecycle slice.
3. Add worker-scoped fixtures as a complete lifecycle slice, including `isolate: true`, `isolate: false`, watch rebuilds, bounded cleanup, teardown ordering, logging, and finalized run errors.

The first follow-up will intentionally avoid the worker-pool and watch lifecycle changes from this PR.

PR #1677, opened 2026-08-05
feat(core): support project-level silent config

fi3ework:
## Summary

- Allow `silent` in project configs and expose each project's effective value through `getRstestConfig()`.
- Route browser console handling through per-project controllers so projects can use different `silent` values and `passed-only` buffers cannot collide when projects share test paths.
- Add Node and browser regressions covering independent silent and non-silent projects.

## Checklist

- [x] Documentation updated (or not required; project config scoping is already documented).

### DIFF
--- a/e2e/browser-mode/console.test.ts
+++ b/e2e/browser-mode/console.test.ts
@@ -66,6 +66,16 @@ describe('browser mode - console forwarding', () => {
     },
   );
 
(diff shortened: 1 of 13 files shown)


fi3ework:
handoff to @9aoy

9aoy (approved):

PR #1695, opened 2026-08-07
perf(core): preserve binary assets across workers

9aoy:
## Summary

This PR keeps compiled asset contents in byte form across the host, pool, and worker boundaries, and decodes only at semantic text-consumption points. This avoids UTF-8 corruption/expansion for binary assets while preserving text-module behavior.

- Node fork/thread transports retain byte content; Bun uses a JSON-safe representation when required.
- Virtual filesystem reads return exact binary bytes.
- The asset-heavy fixture used below contains about **16.9 MB of non-UTF-8 resources**. The previously observed **~30.1 MB** UTF-8 re-count reflects decoding expansion/corruption risk, not exact IPC wire size.

### Performance evidence

Same source, same test set, same Node version, and 5-run warm medians:

| Scenario | Metric | UTF-8 string | Buffer-first | Change |
| --- | --- | ---: | ---: | ---: |
| Text-only transport (`emitAssets: false`) | CLI wall | 2.13s | 1.96s | **-8%** |
|  | Rstest Duration | 841ms | 748ms | **-11%** |
|  | Peak RSS | 1,042.5MB | 1,014.1MB | **-2.7%** |
| Asset transport enabled (`emitAssets: true`) | CLI wall | 3.36s | 2.24s | **-33%** |
|  | Rstest Duration | 1,560ms | 909ms | **-42%** |
|  | Peak RSS | 1,156.1MB | 1,083.1MB | **-6.3%** |

The binary-heavy case shows a material benefit, while the text-only case remains broadly stable.

For an additional text-only Rsbuild compatibility suite (56 files, 378 tests), the measured change was 1.42s → 1.35s CLI wall (-4.9%) and 318MB → 313MB peak RSS (-1.6%), which is not a material overall difference.

## Checklist

### DIFF
--- a/e2e/federation/fixtures/basic/index.test.ts
+++ b/e2e/federation/fixtures/basic/index.test.ts
@@ -1,4 +1,6 @@
+import { resolve } from 'node:path';
 import { expect, it } from '@rstest/core';
+import imagePath from '../../../../examples/federation/component-app/src/MF.jpeg';
 
 it('should expose the federation flag and dynamic import fallback', async () => {
   expect((globalThis as any).__rstest_federation__).toBe(true);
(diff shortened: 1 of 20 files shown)


9aoy:
@codex review

PR #1757, opened 2026-08-28
fix(vscode): make icon setup cross-platform

### LINKED ISSUE #1756: [Bug]: VS Code preinstall redownloads an existing icon on Windows
## Version

```sh
System:
  OS: Microsoft Windows 10 Pro N 10.0.19045 (64-bit)
  Node: v24.14.1
  pnpm: 11.24.0
  Rstest repository: 95172d0ac0266025d9f03926ad2834a54e585c22 (main)
```

## Details

The VS Code package's `preinstall` script uses the POSIX expression `[ -f icon.png ]` as its existence guard. pnpm runs that script through `cmd.exe` on Windows, where `[` is not a command. The failed guard therefore reaches the `curl` fallback and downloads the 28,214-byte icon on every install, even when `packages/vscode/icon.png` already exists.

Expected behavior: an existing icon should make the preinstall step a no-op.

Actual output includes:

```text
packages/vscode preinstall: '[' is not recognized as an internal or external command,
packages/vscode preinstall: operable program or batch file.
packages/vscode preinstall: 100 28214  100 28214
```

This creates an unnecessary CDN dependency for normal Windows workspace installation.

## Reproduce link

https://github.com/web-infra-dev/rstest/blob/95172d0ac0266025d9f03926ad2834a54e585c22/packages/vscode/package.json#L21

## Reproduce Steps

1. Clone the repository on Windows and check out `main`.
2. Ensure `packages/vscode/icon.png` exists (a prior install creates it).
3. Run `pnpm install --frozen-lockfile`.
4. Observe that the POSIX guard fails and `curl` downloads the icon again.

dajiaohuang:
## Summary

- Windows runs the VS Code package lifecycle through `cmd.exe`, so the POSIX icon guard fails and redownloads an existing icon.
- Replace the shell expression with a dependency-free Node helper and cover both the existing-icon and missing-icon paths.
- This only affects VS Code extension asset setup; browser mode, adapters, public APIs, and documentation are unaffected.

## Related Links

- Fixes #1756

## Checklist

### DIFF
--- a/packages/vscode/package.json
+++ b/packages/vscode/package.json
@@ -18,7 +18,7 @@
   "scripts": {
     "build": "rslib build && npx prettier ./LICENSE.md --write",
     "build:local": "cross-env SOURCEMAP=true rslib build",
-    "preinstall": "[ -f icon.png ] || curl https://assets.rspack.rs/rstest/rstest-logo-512x512.png --output icon.png",
+    "preinstall": "node ./scripts/ensureIcon.cjs",
     "lint": "rslint .",
     "package:vsce": "npm run build && vsce package",
     "test": "npm run test:unit && npm run test:e2e",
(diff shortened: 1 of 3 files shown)


9aoy:
👍

Issue #767, opened 2025-12-14
[Bug]: Rstest is slower than Jest


dmitriybo95:
### Version

```sh
System:
    OS: Linux 6.1 Debian GNU/Linux 12 (bookworm) 12 (bookworm)
    CPU: (2) x64 AMD EPYC
    Memory: 2.55 GB / 4.01 GB
    Container: Yes
    Shell: Unknown
  npmPackages:
    @rstest/core: ^0.7.2 => 0.7.2
```

### Details

On more complex React projects with many imports and using scss, rstest becomes slower than jest.
I've attached an example with two identical tests.
Average test execution time:
jest: 1.125s
rstest: 1.15s - 1.6s

### Reproduce link

https://codesandbox.io/p/devbox/92swfz

### Reproduce Steps

1. go to https://codesandbox.io/p/devbox/92swfz
2. fork
3. run the commands `npm run test:jest` and `npm run test:rstest` several times.

9aoy:
In your Jest configuration, you have the following setup:

<img width="531" height="67" alt="Image" src="https://github.com/user-attachments/assets/a496879f-7576-42e6-9567-c114ba25a9fc" />

You can achieve the same effect in rstest with the configuration below. We will provide an equivalent configuration to Jest's `moduleNameMapper` in the future.

```ts
export default defineConfig({
  tools: {
    rspack: (config, { rspack}) => {
      config.plugins.push(
              new rspack.NormalModuleReplacementPlugin(
               /\.(css|less|scss)/,
             "identity-obj-proxy"
            )
      );
    }
  }
});
```

ramonvillain-st:
We are planning to migrate a large Jest suite to Rstest. On a test branch with the full suite running on Rstest, we see the same problem as this issue on our CI. Sharing our numbers in case it helps.

**Project size:** React + jsdom, ~2,400 test files, ~20,000 tests. Jest uses `moduleNameMapper` to replace all CSS/SCSS with `identity-obj-proxy`.

**CI results (Linux container, Jenkins, full suite):**

| Runner | Time |
| --- | --- |
| Jest | ~3 min |
| Rstest | ~10 min |

**What `--trace` shows (234-file sample):**

| phase | share |
| --- | --- |
| tests | 38% |
| envSetup | 37% |
| setupFiles | 14% |
| collect | 9% |

[...]

PR #1823, opened 2026-09-09
fix(core): support VM constructors in expect.any

Gobd:
## Motivation

VM pools execute test files in a separate JavaScript realm. The constructors passed by tests to expect.any therefore differ by identity from the host-realm built-ins that @vitest/expect special-cases, causing primitive asymmetric matches such as expect.any(String) to fail under vmThreads and vmForks.

Related to #767.

## Changes

Normalize VM-realm primitive constructors to their host equivalents before delegating to the existing asymmetric matcher implementation.

Add focused runtime coverage plus node-pool e2e coverage that runs under threads, vmThreads, and vmForks.

### DIFF
--- a/packages/core/src/runtime/api/expect.ts
+++ b/packages/core/src/runtime/api/expect.ts
@@ -223,14 +223,15 @@ export function createExpect({
     for (const name of ANY_PRIMITIVE_CONSTRUCTOR_NAMES) {
       if (constructor === runtimeGlobal?.[name]) {
         const matcher = Reflect.apply(any, expect, [globalThis[name]]);
+        if (name === 'Function' || name === 'Object') {
+          return matcher;
+        }
         const asymmetricMatch = matcher.asymmetricMatch.bind(matcher);
         matcher.asymmetricMatch = (value: unknown) =>
           asymmetricMatch(value) ||
-          Reflect.apply(
-            Function.prototype[Symbol.hasInstance],
-            constructor,
-            [value],
-          );
+          Reflect.apply(Function.prototype[Symbol.hasInstance], constructor, [
+            value,
+          ]);
         return matcher;
       }
     }
--- a/packages/core/src/runtime/api/expect.ts
+++ b/packages/core/src/runtime/api/expect.ts
@@ -59,6 +59,16 @@ const defaultChaiConfig: ChaiConfig = {
   truncateThreshold: chaiConfig.truncateThreshold,
 };
 
+const ANY_PRIMITIVE_CONSTRUCTOR_NAMES = [
+  'String',
+  'Number',
+  'Function',
+  'Boolean',
+  'BigInt',
+  'Symbol',
+  'Object',
+] as const;
+
 export function setupChaiConfig(config: ChaiConfig = {}): void {
   Object.assign(chaiConfig, defaultChaiConfig, config);
 }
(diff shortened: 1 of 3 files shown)


packages/core/src/runtime/api/expect.ts
```diff
@@ -207,6 +217,17 @@ export function createExpect({
   Object.assign(expect, chaiExpect);
   Object.assign(expect, (globalThis as any)[ASYMMETRIC_MATCHERS_OBJECT]);
 
+  const any = expect.any;
+  expect.any = (constructor) => {
+    const runtimeGlobal = fileContext().runtimeGlobal;
+    for (const name of ANY_PRIMITIVE_CONSTRUCTOR_NAMES) {
+      if (constructor === runtimeGlobal?.[name]) {
+        return Reflect.apply(any, expect, [globalThis[name]]);
```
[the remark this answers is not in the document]
Gobd: Good catch—fixed in 124a0f6d. The matcher still delegates primitive handling to the host-realm constructor, then falls back to the original VM constructor’s default Symbol.hasInstance behavior for boxed values. Regression coverage now includes boxed String, Number, Boolean, BigInt, and Symbol values plus a String subclass. Focused tests (17/17), core build/lint/typecheck, and node-pool e2e across threads, vmThreads, and vmForks all pass.

packages/core/src/runtime/api/expect.ts
```diff
@@ -207,6 +217,26 @@ export function createExpect({
...
+      if (constructor === runtimeGlobal?.[name]) {
+        const matcher = Reflect.apply(any, expect, [globalThis[name]]);
+        const asymmetricMatch = matcher.asymmetricMatch.bind(matcher);
+        matcher.asymmetricMatch = (value: unknown) =>
+          asymmetricMatch(value) ||
+          Reflect.apply(
+            Function.prototype[Symbol.hasInstance],
+            constructor,
+            [value],
+          );
```
9aoy: **[P2] Preserve the existing Object and Function matching semantics**

Applying this `instanceof` fallback to every constructor broadens the existing matchers: `expect.any(Object)` now accepts functions, and `expect.any(Function)` accepts non-callable objects such as `Object.create(Function.prototype)`. The normal threads pool also sets `runtimeGlobal` to `globalThis`, so this introduces a regression outside VM pools as well.

```ts
expect(() => {}).not.toEqual(expect.any(Object));
expect(Object.create(Function.prototype)).not.toEqual(expect.any(Function));
```

I reproduced both failures using the built CLI under `threads`, `vmThreads`, and `vmForks`. Removing this PR's implementation change and rebuilding makes both assertions pass under `threads`.

Could we restrict the fallback to constructors that need boxed-primitive compatibility, preserve the existing `Object`/`Function` matcher semantics, and add these negative regression cases?

9aoy (commented):
Found a matching-semantics regression in the new instanceof fallback; details and reproduction below.

packages/core/src/runtime/api/expect.ts
Gobd: Good catch—fixed in 27e10ddd. The Symbol.hasInstance fallback is now limited to boxed primitive constructors; Object and Function continue using Vitest’s existing matcher semantics. I also added negative unit and node-pool e2e regressions for functions vs. expect.any(Object) and Object.create(Function.prototype) vs. expect.any(Function). The e2e cases failed before the fix under threads, vmThreads, and vmForks, and pass with the fix.

9aoy (approved):

9aoy:
👍

PR #1830, opened 2026-09-10
fix(core): handle cross-realm constructors in toStrictEqual

ramon-villain:
## Motivation

Under `vmThreads` and `vmForks`, `toStrictEqual` fails on values that come from another realm, with "Compared values have no visual difference". Host APIs create such values: `URLSearchParams#getAll()` returns a host `Array`, and `TextEncoder#encodeInto()` returns a host `Object`. On the 2725-file suite from #767, on `73725ba`, this is the last cross-realm class left: 12 files and 65 tests. The same files pass under `forks`.

```ts
it('values from the host realm', () => {
  expect(new URLSearchParams('a=1').getAll('a')).toStrictEqual(['1']);
  expect(new TextEncoder().encodeInto('a', new Uint8Array(1))).toStrictEqual({
    read: 1,
    written: 1,
  });
});
```

`@vitest/expect`'s `typeEquality` returns `false` when `a.constructor !== b.constructor`, and a value from another realm has different built-in constructors. Jest treats two arrays as the same type (jestjs/jest#2549), and two native built-ins with the same constructor name as the same type (jestjs/jest#15959). User classes still need identity. The gap also exists outside VM pools: `expect(vm.runInNewContext("['1']")).toStrictEqual(['1'])` fails under `forks`.

## Changes

- Add a `CrossRealmToStrictEqual` Chai plugin in `runtime/api/expect.ts`, next to `CrossRealmToThrow`. It redefines `toStrictEqual` the way `JestChaiExpect` does, with the same testers from `@vitest/expect`, but with a `typeEquality` that follows Jest's two cross-realm rules. It wraps the matcher with `wrapAssertion`, so `expect.soft` keeps working. No patch of `@vitest/expect` is needed.
- Add a unit test in `tests/runtime/api/expect.test.ts`. A cross-realm array and a cross-realm plain object strictly equal their literals. The matcher stays strict: a cross-realm sparse array, a same-name user class from another realm, a class instance, and an `undefined` property still fail. Each assertion fails when its part of the matcher is removed.
- Extend the node-pool e2e fixture, which runs on `threads`, `vmThreads` and `vmForks`, with both host-realm values.

Not covered: Jest applies the same constructor-name rule inside `iterableEquality`, which covers cross-realm `Map` and `Set`. In `@vitest/expect`, `toEqual`, `toStrictEqual`, `toMatchObject` and the call-argument matchers call that module-private tester directly, so it belongs in Vitest's `jest-utils.ts`, not in a plugin here.

### DIFF
--- a/e2e/node-pool/fixtures/test/basic.test.ts
+++ b/e2e/node-pool/fixtures/test/basic.test.ts
@@ -34,6 +34,14 @@ describe('node pool - basic', () => {
     ]);
   });
 
+  it('strictly equals built-ins created in the host realm', () => {
+    expect(new URLSearchParams('a=1').getAll('a')).toStrictEqual(['1']);
+    expect(new TextEncoder().encodeInto('a', new Uint8Array(1))).toStrictEqual({
+      read: 1,
+      written: 1,
+    });
+  });
+
   it('matches VM-realm constructors in asymmetric matchers', () => {
     expect('value').toEqual(expect.any(String));
     expect(42).toEqual(expect.any(Number));
--- a/packages/core/src/runtime/api/expect.ts
+++ b/packages/core/src/runtime/api/expect.ts
@@ -20,15 +20,22 @@
 import {
   ASYMMETRIC_MATCHERS_OBJECT,
   addCustomEqualityTesters,
+  arrayBufferEquality,
   ChaiStyleAssertions,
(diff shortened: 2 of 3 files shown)


9aoy (approved):

9aoy:
👍

PR #1851, opened 2026-09-14
test(mock): reproduce manual mock ignored for workspace-linked package

### LINKED ISSUE #1850: [Bug]: Manual mock colocated inside a workspace/node_modules package's own __mocks__ is ignored (falls back to automock)
### Version

```sh
@rstest/core: 0.11.11 / 0.11.12
```

### Details

**A manual mock (`__mocks__`) that lives inside the *mocked* package's own directory (Jest's documented convention for mocking node_modules / workspace packages) is silently ignored by Rstest, which falls back to full automocking instead.**

Rstest's manual-mock lookup only ever checks a single fixed directory: `<rootPath>/__mocks__`, where `rootPath` is the root of the package **under test** (the consumer). It never looks inside the resolved, real directory of the module being mocked.

Jest's equivalent lookup is richer: for a bare `jest.mock("some-pkg")` call (no factory), Jest resolves `some-pkg`, follows any symlink (e.g. a package-manager workspace link) to its real location, and checks for a manual mock colocated there, e.g. `packages/some-pkg/__mocks__/index.js`.

This pattern — a shared package shipping its own hand-written mock next to its real implementation (very common for design-system / utility packages, e.g. Fluent UI–style `makeStyles` helpers) — is idiomatic in Jest-based monorepos. Under Rstest, `rs.mock("some-pkg")` (or `jest.mock` via the Jest compat layer) silently falls back to automocking every export of `some-pkg` instead of using that manual mock, which turns any mocked function export into a stub that returns `undefined` by default. Any consumer code that treats the return value as a real function (e.g. `const useStyles = makeStyles(...); useStyles();`) then crashes with `"... is not a function"`.

I found this while migrating a large monorepo (Jest → Rstest side-by-side) — it is the single largest remaining source of failures in that migration (hundreds of "X is not a function" errors traced back to this one root cause).

I searched existing issues and didn't find this exact scenario reported. Related-but-distinct issues:
- #1479 (merged docs PR) confirms the *intended* `__mocks__` → automock fallback design, but doesn't cover the mocked-package's-own-`__mocks__` case.
- #508 (merged) fixed `manualMockRoot` to use `project.rootPath` in multi-project configs — about *which* rootPath, not about checking the mocked package's own directory.
- #1456 is a related-but-different externalization/mock-propagation bug for externalized `node_modules` deps.

### Reproduce Steps

Minimal reproduction (2 linked workspace packages + jest/rstest diff):

`styles-lib/package.json`
```json
{ "name": "styles-lib", "version": "1.0.0", "main": "index.js" }
```

`styles-lib/index.js` (real implementation)
```js
exports.makeStyles = function makeStyles(definitionsFn) {
  return function useStyles() {
    return definitionsFn();
  };
};
```

`styles-lib/__mocks__/index.js` (manual mock, colocated inside styles-lib's OWN directory)
```js
exports.makeStyles = function makeStyles() {
  return function useStyles() {
    return { mocked: true };
  };
};
```

`consumer/component.js`
```js
const { makeStyles } = require("styles-lib");
const useCardStyles = makeStyles(() => ({ real: true }));

function getCardStyles() {
  return useCardStyles(); // throws if useCardStyles is undefined
}

module.exports = { getCardStyles };
```

`consumer/component.jest.test.js` — **passes** under Jest
```js
const { getCardStyles } = require("./component");
jest.mock("styles-lib"); // bare call, no factory
test("uses the manual mock", () => {
  expect(getCardStyles()).toEqual({ mocked: true });
});
```

`consumer/component.rstest.test.js` — **fails** under Rstest
```js
import { expect, it, rs } from "@rstest/core";
const { getCardStyles } = require("./component");
rs.mock("styles-lib"); // same bare call, no factory
it("uses the manual mock", () => {
  expect(getCardStyles()).toEqual({ mocked: true });
});
```

`styles-lib` and `consumer` are linked via npm/yarn/pnpm workspaces (so `node_modules/styles-lib` is a real symlink to `styles-lib/`, matching a real monorepo).

1. `npx jest --ci` in `consumer/` → ✅ passes (`getCardStyles()` returns `{ mocked: true }`).
2. `npx rstest run` in the same `consumer/` → ❌ fails:
   ```
   TypeError: useCardStyles is not a function
       at getCardStyles (component.js:...)
   ```
   because `makeStyles` was automocked (Rstest never found `styles-lib/__mocks__/index.js`) and returned `undefined`.

### I also adapted this into an `e2e/mock` fixture matching this repo's own test conventions

I cloned this repo and added:
- `e2e/mock/fixtures/manualMockWorkspacePackage/styles-lib/` — the real + manually-mocked workspace package (linked via `workspace:*`)
- `e2e/mock/fixtures/manualMockWorkspacePackage/consumer/` — the consumer package with `rstest.config.mts` and `component.test.ts` (bare `rs.mock("...")`)
- `e2e/mock/tests/manualMockWorkspacePackage.test.ts` — wires the fixture into the harness via `runRstestCli` + `expectExecFailed()`, so it currently passes (documenting the bug) and can be flipped to `expectExecSuccess()` once fixed

I verified all of this against a real build of this repo:
- `pnpm install --filter ...` + `pnpm --filter @rstest/core run build` to build core from source
- `rstest run` directly inside the fixture's `consumer/` package → fails with `useCardStyles is not a function`
- `cd e2e && npx rstest run mock/tests/manualMockWorkspacePackage.test.ts` (the real e2e harness) → passes, correctly asserting the CLI run currently fails

Happy to open this as a PR (fixture + failing-test) if that's useful — let me know and I'll push a branch.

### Expected

`rs.mock("styles-lib")` (or the Jest-compat `jest.mock("styles-lib")`) should find and use the manual mock colocated at `styles-lib/__mocks__/index.js`, matching Jest's behavior, instead of silently falling back to automocking.

### Actual

The manual mock is ignored. `makeStyles` is automocked (returns `undefined` by default), and any consumer code that calls the mocked function's return value as a function throws `"... is not a function"`.

kelleyma49:
Adds a failing-by-design e2e fixture + test reproducing #1850.

## What

- `e2e/mock/fixtures/manualMockWorkspacePackage/styles-lib` — a workspace package that ships a manual mock colocated at its own `__mocks__/index.js` (Jest's documented convention for mocking node_modules / workspace packages).
- `e2e/mock/fixtures/manualMockWorkspacePackage/consumer` — a workspace-linked consumer that calls `rs.mock("...")` (bare, no factory) on `styles-lib`.
- `e2e/mock/tests/manualMockWorkspacePackage.test.ts` — wires the fixture into the harness via `runRstestCli` + `expectExecFailed()`.

## Why it's written to expect failure

Rstest's manual-mock lookup only checks `<rootPath-of-the-package-under-test>/__mocks__`, so it never finds `styles-lib`'s own colocated manual mock and falls back to automocking `makeStyles`, which then returns `undefined`. The consumer crashes with `useCardStyles is not a function`. Jest, by contrast, resolves the real (symlinked) module path and correctly finds the colocated mock.

`expectExecFailed()` documents this current (buggy) behavior as an explicit e2e test. Once manual-mock resolution is fixed to follow the resolved module's real path (matching Jest), this one line should be flipped to `expectExecSuccess()` — that flip is intended to be the regression test for the eventual fix.

## Verification

- `pnpm install --filter ...` + `pnpm --filter @rstest/core run build`, then `rstest run` directly inside `consumer/` → fails with `useCardStyles is not a function`.
- `cd e2e && npx rstest run mock/tests/manualMockWorkspacePackage.test.ts` (the real e2e harness) → passes, correctly asserting the CLI run currently fails.

Closes #1850 (as a repro; the underlying manual-mock resolution bug is still open).

## Checklist


9aoy (approved):

9aoy:
@kelleyma49  Thanks for reporting this issue and contributing the regression test! The issue has been fixed, and the test now passes. The fix will be included in Rstest 0.12.

PR #1863, opened 2026-09-18
fix(core): drain worker IPC writes before async leak checks

9aoy:
## Motivation

Async leak detection can sample Rstest's own fork-worker `process.send()` `WRITEWRAP` before the IPC write has completed, causing a false positive in otherwise clean tests. The reported zlib failure was this worker IPC race; the zlib stream itself is already covered by the existing `ZLIB` handling.

## Changes

- track outstanding local `process.send()` callbacks inside `ForksChannel`
- drain pending RPC writes after the file run and before async leak sampling
- keep runner hooks and RPC responses asynchronous; only the local IPC write completion is awaited
- settle pending writes when the IPC channel disconnects, while thread workers remain a no-op
- keep `WRITEWRAP` detectable so user-created IPC leaks are not globally hidden
- add unit coverage for write draining and disconnect behavior

## Validation

- core lint and build
- root typecheck
- fork channel and RPC unit tests
- async leak e2e in ESM and CommonJS output modes
- targeted zlib async leak e2e passed 30 consecutive runs
- `pnpm check-unused`


9aoy:
@codex review

PR #1864, opened 2026-09-18
perf(core): cache test entry compilation in VM pools

Gobd:
## Motivation

`vmThreads` already reuses bounded V8 compilation metadata for setup files, but test entries do not enable the same cache. Because the entry loader propagates this option into shared chunks, large isolated suites repeatedly parse the same ESM dependency chunks in fresh VM contexts.

On a 290-file / 1,536-test jsdom shard, tracing attributed 20.39s of aggregate collection time to constructing 3,887 dependency `SourceTextModule`s. Enabling the existing cache reduced that work to 2.66s while preserving a fresh context, module graph, and evaluation for every file.

## Changes

Enable `cacheCompilation` when loading test entries in VM pools, matching setup-file loading. This reuses only V8 parse/compile metadata through the existing bounded worker cache; module instances and state remain file-isolated.

Paired untraced runs showed a conservative ~6% wall-time improvement. On current `main`, three interleaved controls were 19.1s, 19.1s, and 19.2s; the final patched pair was 17.4s and 17.3s. An earlier patched run was a 25.1s cold/machine outlier, followed by 18.8s, so the PR does not claim the best-case 9.3% universally.

Validation:

- core build
- focused compilation-cache and worker unit tests: 13 passed
- node-pool E2E, including vmThreads isolation: 7 passed
- pre-push package builds, typecheck, formatting, spelling, and dependency consistency checks

Browser mode and adapters are unaffected because this option is consumed only by the Node VM loaders.


9aoy (approved):

9aoy:
👍

